PR #21, opened 2020-08-26
Actually fix multiplication by infinity

LLFourn:
Apparently github actions gave a false negative and said it was ok when it wasn't.


LLFourn:
No. I am wrong. Travis is right.

PR #29, opened 2020-09-23
Sigma fun

LLFourn:
This introduces a new crate for composing sigma protocols using textbook `and`, `or`, `eq` composition. Hopefully it will eventually have threshold (m of n) and `neq` composition too.
It also provides sane and secure Fiat-Shamir transformation to a non-interactive proof system.

The idea is that you could take a specification like `eq(2,DL-secp256k1)` and compile it with a macro:
```rust
// create a proof for showing the DLOG of two points are the same
let proof_system = compile_sigma_proof_system!(eq(2,DL-secp256k1));
let proof = proof_system.prove(((G,xG,),(H, xH)), x);
```
TODO:
- `all` composition so you can and compose the same proof many times e.g. `all(256, DL-secp256k1)`
- the actual macro that does the compilation

I will probably merge this when I have done `all` and implemented the cross-curve DLEQ proof using it.


LLFourn:
This is what this crate needs before publishing:

- [ ] Write docs
- [ ] serde
- [ ] Vec security issues
- [ ] Avoid mutability in `Transcript` API

sigma_fun/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "sigma_fun"
+version = "0.1.0"
+authors = ["LLFourn <lloyd.fourn@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+generic-array = "0.14"
```
LLFourn, line 10: digest re-exports generic-array so let's just use that.

sigma_fun/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+pub use all::All;
+pub mod ext;
+mod transcript;
+pub use transcript::*;
+
+pub trait Sigma {
+    type Witness;
+    type Statement;
+    type AnnounceSecret;
+    type Announce: core::cmp::Eq + core::fmt::Debug;
```
LLFourn, line 44: s/Announce/Announcement/g

sigma_fun/src/ed25519.rs
```diff
@@ -0,0 +1,203 @@
...
+        hash.update(announcement.compress().as_bytes())
+    }
+
+    fn hash_witness<H: Digest>(&self, hash: &mut H, witness: &Self::Witness) {
+        hash.update(witness.to_bytes().as_ref())
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DLBP<L> {
```
LLFourn, line 97: s/DLBP/DLG/ so that BP is not confused with "bulletproof".

sigma_fun/src/ext/dl_secp256k1_ed25519_eq.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///
+    /// - If the secret is larger than 2^253 -1
+    /// - If the secret is 0
+    pub fn prove(
+        &self,
+        secret: &ScalarQ,
+        rng: &mut (impl CryptoRng + RngCore),
+    ) -> CrossCurveDLEQProof {
+        // Must be a 252 bit ed25519 key
+        assert!(secret.as_bytes()[31] & 0x20 == 0);
```
luckysori, line 88: I'm a real newbie when it comes to bitwise operations, but I think that this only proves that the scalar is 255 bits, not 252.

PR #30, opened 2020-09-29
Derive default for Scalar

luckysori:
There may be a reason why you are not already deriving `Default` for `Scalar`, but I can't find it.

### DIFF
--- a/secp256kfun/src/scalar.rs
+++ b/secp256kfun/src/scalar.rs
@@ -290,6 +290,24 @@ impl HashInto for Scalar {
     }
 }
 
+impl<S> Default for Scalar<S, Zero>
+where
+    S: Secrecy,
+{
+    fn default() -> Self {
+        Scalar::zero().mark::<S>()
+    }
+}
+
+impl<S> Default for Scalar<S, NonZero>
+where
+    S: Secrecy,
+{
+    fn default() -> Self {
+        Scalar::one().mark::<S>()
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;


LLFourn:
There is a problem with this:
```
let non_zero_scalar = Scalar::<_, NonZero>::default();
```
what should this do? It probably results in `0` which is a violation of the guarantee.
If it gets a default it has to be manually implemented for both cases.
This would be fine but I don't see a strong reason to have a `Default` for Scalar?

luckysori:
> If it gets a default it has to be manually implemented for both cases.

You're right!

> This would be fine but I don't see a strong reason to have a `Default` for Scalar?

There's no strong reason other than convenience. When using `generic-array` in `cross-curve-dleq` it was concise to generate the empty generic array of scalars using a default method https://github.com/comit-network/cross-curve-dleq/pull/2/commits/81a33948aded64ab37e1bbba5c04e7a441417f53#diff-b4aea3e418ccdb71239b96952d9cddb6R215

thomaseizinger:
> If it gets a default it has to be manually implemented for both cases.

It could also just be implemented for one where you specify certain generic types, doesn't have to be implemented for both :)

LLFourn:
@luckysori

1. The idea I had in mind was to use the functional api `.map/.zip` of generic array which allows you to both preserve type safety of the length and avoid having to first initialize and then set data (i.e `Default`)
2. It turns out in this case this is a bad idea (sorry) -- I tried this [here](https://github.com/LLFourn/secp256kfun/pull/29/commits/8edb9c8c0c4de8b825f1725f156e54f26a6ba346#diff-e848b46f45e2cd56b3c1ff7d99dc9b91R94). It led to a stack overflow. I am not sure why. The amount of data on the stack is large in this case but not nearly enough to overflow the usual 4-8mb limit. GenericArray uses a lot of unsafe code to achieve its API and I guess something must be going wrong.
3. If you really want to do this you can always create a generic array like: 
```rust
GenericArray::<Scalar,U256>::from_exact_iter(core::iter::repeat(Scalar::one()).take(256))
```

Having said that I don't mind having a `Default` for Scalar where it's 0 for `Zero` and 1 for `NonZero` if you want to change the PR to that.

(side note let's have call soon to discuss this proof further).

luckysori:
> @luckysori
> 
>     1. The idea I had in mind was to use the functional api `.map/.zip` of generic array which allows you to both preserve type safety of the length and avoid having to first initialize and then set data (i.e `Default`)
> 
>     2. It turns out in this case this is a bad idea (sorry) -- I tried this [here](https://github.com/LLFourn/secp256kfun/pull/29/commits/8edb9c8c0c4de8b825f1725f156e54f26a6ba346#diff-e848b46f45e2cd56b3c1ff7d99dc9b91R94). It led to a stack overflow. I am not sure why. The amount of data on the stack is large in this case but not nearly enough to overflow the usual 4-8mb limit. GenericArray uses a lot of unsafe code to achieve its API and I guess something must be going wrong.
> 
>     3. If you really want to do this you can always create a generic array like:
> 
> 
> ```rust
> GenericArray::<Scalar,U256>::from_exact_iter(core::iter::repeat(Scalar::one()).take(256))
> ```
> 
> Having said that I don't mind having a `Default` for Scalar where it's 0 for `Zero` and 1 for `NonZero` if you want to change the PR to that.

Thank you for the thorough explanation! I think I will change this PR to do it properly, since using `Default` to initialise `GenericArray`s is a little bit terser.

PR #37, opened 2020-11-04
Replace ordinary tests with proptests

luckysori:
I tried to modify all the tests that would benefit from this approach. Let me know if I missed some.

I added a `proptest` feature flag to `secp256kfun` so that we can define the functions which generate the test data in one place and a consumer of the library (such as the `sigma_fun` crate) can use them to write property-based tests.

I configured the tests in `dl_secp256k1_ed25519_eq.rs` with `#![proptest_config(ProptestConfig::with_cases(10))]`, because these are slow tests (2-4 seconds). You may want to only run them once as opposed to 10 times if you care about CI speed.

### DIFF
--- a/sigma_fun/Cargo.toml
+++ b/sigma_fun/Cargo.toml
@@ -9,13 +9,14 @@ edition = "2018"
 [dependencies]
 generic-array = "0.14"
 digest = "0.9"
-secp256kfun = { path = "../secp256kfun", version = "^0.3.2-alpha.0", default-features = false, optional = true, features = ["proptest"] }
+secp256kfun = { path = "../secp256kfun", version = "^0.3.2-alpha.0", default-features = false, optional = true }
 curve25519-dalek = { version = "3", default-features = false, optional = true, features = ["u64_backend"] }
 rand_chacha = "0.2"
 serde_crate = { package = "serde", version = "1.0", optional = true, default-features = false, features = ["derive"] }
 proptest = "0.10"
 
 [dev-dependencies]
+secp256kfun = { path = "../secp256kfun", version = "^0.3.2-alpha.0", default-features = false, features = ["proptest"] }
 rand = "0.7"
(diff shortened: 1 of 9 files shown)


sigma_fun/Cargo.toml
```diff
@@ -9,10 +9,11 @@ edition = "2018"
 [dependencies]
 generic-array = "0.14"
 digest = "0.9"
-secp256kfun = { path = "../secp256kfun", version = "^0.3.2-alpha.0", default-features = false, optional = true }
+secp256kfun = { path = "../secp256kfun", version = "^0.3.2-alpha.0", default-features = false, optional = true, features = ["proptest"] }
```
thomaseizinger, line 12: You should be able to move this to `dev-dependencies` so that this feature is only enabled during testing.

LLFourn:
@luckysori I thought that through proptesting you were able to capture that https://github.com/LLFourn/secp256kfun/pull/29/files#diff-bd3a9ec9d692ecc24a138f50f244dc99486181b441556dd6ffb607ae5d5442edR93 is checking the wrong bit?

sigma_fun/Cargo.toml
```diff
@@ -13,8 +13,10 @@ secp256kfun = { path = "../secp256kfun", version = "^0.3.2-alpha.0", default-fea
 curve25519-dalek = { version = "3", default-features = false, optional = true, features = ["u64_backend"] }
 rand_chacha = "0.2"
 serde_crate = { package = "serde", version = "1.0", optional = true, default-features = false, features = ["derive"] }
+proptest = "0.10"
```
LLFourn, line 16: ❓ is it possible to just use the re-exported one from secp256kfun?

LLFourn:
Thanks @luckysori!

I will fix the bug in the main PR and see if I can stop `proptest` being a core dependency of `sigma_fun`.

PR #57, opened 2021-02-19
Implement subtle::ConditionallySelectable for Point

LLFourn:
So we can do constant time selection for them in the cross curve dleq proof.

@luckysori this fixes the last remaining issue for the proof that I am aware of.


luckysori:
Good work!

Issue #64, opened 2021-05-04
Move away from nightly?


thomaseizinger:
Are there any plans to move away from Rust nightly?

Specialization doesn't look like it is going to be stabilized any time soon.

LLFourn:
It is very easy to remove specialization and just put the default (slow) implementation of each primitive as the only one and then enable the faster variants with a `nightly` feature.

Would that be of help?

thomaseizinger:
That would be really nice as secpfun is the only dep forcing us on nightly atm!

I'll see to send a PR for that!

LLFourn:
So to be clear the things that need to be done are:

1) from the base implementation (the one where everything is marked default for each impl chain) remove `default` i.e.
get rid of them for here unless `nightly` feature enabled from here

https://github.com/LLFourn/secp256kfun/blob/26a0f272d722257ae9d49d764fef609f1e0a92dc/secp256kfun/src/op.rs#L93-L108

I suggest trying to make a macro that parses a list of functions and outputs two versions (one with `not(feature = nightly)` and the other with `feature = nightly`.

2) for each specializing impl just put `#[cfg(feature = nightly)]` on top of it
3) Put `README.md` embedding which needs `extended_key_value_attributes` under nightly as well: https://github.com/LLFourn/secp256kfun/blob/26a0f272d722257ae9d49d764fef609f1e0a92dc/secp256kfun/src/lib.rs#L2

this should be all we need to run on stable.

thomaseizinger:
Fixed in #65.

PR #65, opened 2021-05-21
Feature-gate all nightly features behind a nightly feature

thomaseizinger:
Attempting to solve #64.

For details see commit messages.

I've encountered three problems so far:

1. We can't conditionally use `extended_key_value_attributes` on nightly because we already fail during parsing time for an attribute like `#[cfg_attr(feature = "nightly", extended_key_value_attributes)]`. However, the feature is being stabilized with 1.54 which is in 10 weeks. Still quite some time away so not sure how to deal with this. For now I've commented it out to make progress on the more important items.
2. For some reason, rustdoc is mucking around now and can't find `OsRng`? I suspect it has to do with the new feature resolver I activated as part of the patch but I haven't found a solution yet.
3. There seems to be an issue where, once we don't use specialization, the `g_to_and_from_bytes` test fails. If compiled in debug mode, we hit an assertion that the point is not normalized yet. I don't understand the details but you can reproduce the failing test with `cargo +stable test point::test::g_to_and_from_bytes --no-default-features`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,70 +21,31 @@ jobs:
           command: fmt
           args: --all -- --check
 
-  test-nostd:
+  test:
     runs-on: ubuntu-latest
-
+    strategy:
+      matrix:
+        args: [
+            --no-default-features,
+            --no-default-features --features=serde,
+            --no-default-features --features=std,
+            --all-features,
+        ]
+        rust: [nightly, stable]
+        exclude:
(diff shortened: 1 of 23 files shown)


LLFourn (approved):
Awesome work! Thanks for fixing mess as you went along. Looks really good. I've resolved the issues you mentioned and pushed to the branch. I don't mind having to wait until the feature hits stable. Docs will just be more awkward than usual until then.

PR #67, opened 2021-07-23
Musig2 key aggregation

RCasatta:
~~super~~ draft status 

Wanted to try to implement https://github.com/ElementsProject/secp256k1-zkp/blob/master/src/modules/musig/musig-spec.mediawiki

Test vectors are currently passing, however, spec may change https://github.com/ElementsProject/secp256k1-zkp/pull/120#discussion_r676194896

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,6 @@ members = [
     "schnorr_fun",
     "ecdsa_fun",
     "secp256kfun_parity_backend",
-    "sigma_fun"
+    "sigma_fun",
 ]
 resolver = "2"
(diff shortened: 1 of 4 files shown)


LLFourn:
Nice. I think this belongs in `schnorr_fun` eventually.  This would be the first multi-party protocol in the *_fun crates and I have no idea what the API should be like so interested to know if you have some further ideas.  I also need this by November for my own schemes!

RCasatta:
> I think this belongs in schnorr_fun eventually.

Moved into schnorr_fun and removed some TODOs, a little less drafty, however, spec may change https://github.com/ElementsProject/secp256k1-zkp/pull/120#discussion_r676194896

> I have no idea what the API should be like so interested to know if you have some further ideas

At the moment I am trying to get a deeper understanding by playing around to have some ideas

> I also need this by November for my own schemes!

November is coming!

LLFourn:
This is superseded by https://github.com/LLFourn/secp256kfun/pull/80

Thanks for doing this @RCasatta. I used the test case you developed here for the new PR. If you have time to review, please let me know what you think of the API etc.

PR #80, opened 2022-01-12
Implement MuSig2 multisignature scheme

LLFourn:
Sort of compatible with: https://github.com/ElementsProject/secp256k1-zkp/pull/131

relevant new code is in [`musig.rs`](https://github.com/LLFourn/secp256kfun/blob/310c55549ec0571ff2aef51dbee7c79c7a5f1d14/schnorr_fun/src/musig.rs) the rest is just yak shaving.

### DIFF
--- a/schnorr_fun/src/keypair.rs
+++ b/schnorr_fun/src/keypair.rs
@@ -15,7 +15,7 @@ use secp256kfun::{marker::*, Point, Scalar, XOnly};
 /// ```
 ///
 /// [`Schnorr`]: crate::Schnorr
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct KeyPair {
     pub(crate) sk: Scalar,
     pub(crate) pk: XOnly,
--- a/schnorr_fun/src/keypair.rs
+++ b/schnorr_fun/src/keypair.rs
@@ -27,7 +27,7 @@ impl KeyPair {
         &self.sk
     }
 
-    /// Returns a reference to the public key.
+    /// The public key
     pub fn public_key(&self) -> XOnly {
         self.pk
     }
--- a/schnorr_fun/src/keypair.rs
+++ b/schnorr_fun/src/keypair.rs
@@ -28,8 +28,8 @@ impl KeyPair {
     }
 
     /// Returns a reference to the public key.
-    pub fn public_key(&self) -> &XOnly {
-        &self.pk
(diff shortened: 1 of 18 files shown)


schnorr_fun/src/keypair.rs
```diff
@@ -28,8 +28,8 @@ impl KeyPair {
     }
 
     /// Returns a reference to the public key.
-    pub fn public_key(&self) -> &XOnly {
-        &self.pk
+    pub fn public_key(&self) -> XOnly {
```
RCasatta, line 31: Any rationale on why you go this way (by value instead ref?)

schnorr_fun/src/lib.rs
```diff
@@ -2,20 +2,30 @@
 #![no_std]
 #![allow(non_snake_case)]
 #![doc = include_str!("../README.md")]
-#![deny(warnings, missing_docs)]
+#![warn(missing_docs)]
+// #![deny(warnings, missing_docs)]
```
RCasatta, line 6: You may want to restore this?

schnorr_fun/src/musig.rs
```diff
@@ -0,0 +1,1210 @@
...
+//! let session = musig.start_sign_session(&keylist, my_nonces, [p1_nonce, p3_nonce], message).unwrap();
+//! // but since we're using deterministic nonce generation we can just remember the session id.
+//! // You should guarantee that this is not called ever again with the same session id!!!!
+//! let session = musig.start_sign_session_deterministic(&keylist, [p1_nonce, p3_nonce], b"session-id-1337", message).unwrap();
+//! // sign with our (single) local keypair
+//! let my_sig = musig.sign_all(&keylist, &session)[0];
+//! # let _sigs = musig.sign_all(&_keylist, &_session);
+//! # let p1_sig = _sigs[0];
+//! # let p3_sig = _sigs[1];
+//! // recieve p1_sig and p3_sig from somewhere and check they're valid
```
RCasatta, line 49: s/recieve/receive

schnorr_fun/src/keypair.rs
LLFourn: `XOnly` is Copy so I *think* it's conventional to not return references to `Copy` things. On top of that There's no reason to leak the implementation that `keyPair` stores an `XOnly` (it could store a `Point<EvenY>` instead). 

(that comment needs to change though).

PR #84, opened 2022-04-04
FROST Implementation

nickfarrow:
FROST (Flexible Round-Optimized Schnorr Threshold) implementation
* Only handles x-only tweaking (no ordinary tweaks)
* `t-of-n` keygen and signing proptest with random parties, threshold, and signers. Proptest carries out keygen and then uses a random mask to create a set of signers who create a combined signature.
* Docs mostly in synopsis, do we want to add usage docs to particular functions also?
* End-to-end test

This implementation has not yet been made compatible with other FROST implementations or forthcoming specifications.
https://github.com/ElementsProject/secp256k1-zkp/pull/138
https://github.com/isislovecruft/frost-dalek
https://github.com/cfrg/draft-irtf-cfrg-frost

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -220,13 +220,10 @@ impl<SS, H> Frost<SS, H> {
             .ok_or(FirstRoundError::ZeroJointKey)?
             .into_point_with_even_y();
 
-        joint_poly.0[0].conditional_negate(needs_negation);
-
         for poly in &mut point_polys {
             poly.0[0] = poly.0[0].conditional_negate(needs_negation);
         }
-        // TODO ALSO NEGATE JOINT_POLY?
-        joint_poly = PointPoly::combine(point_polys.clone().into_iter());
+        joint_poly.0[0] = joint_poly.0[0].conditional_negate(needs_negation);
 
         let verification_shares = (1..=point_polys.len())
             .map(|i| joint_poly.eval(i as u32).normalize().mark::<NonZero>())
@@ -273,8 +270,8 @@ impl<SS, H> Frost<SS, H> {
 #[derive(Clone, Debug, PartialEq)]
 pub struct SignSession {
     binding_coeff: Scalar,
-    agg_nonces: [Point; 2],
-    group_commitment: Point,
+    nonces_need_negation: bool,
+    agg_nonce: Point<EvenY>,
     challenge: Scalar,
     nonces: HashMap<u32, Nonce>,
 }
@@ -286,10 +283,10 @@ impl<H: Digest<OutputSize = U32> + Clone, CH: Digest<OutputSize = U32> + Clone, 
         // Change to handle multiple my_indexes
         // https://people.maths.ox.ac.uk/trefethen/barycentric.pdf
         // Change to one inverse
-        let k = Scalar::from(my_index as u32);
+        let k = Scalar::from(1 + my_index as u32);
         nonces
             .iter()
-            .map(|(j, _)| *j)
+            .map(|(j, _)| 1 + *j)
             .filter(|j| Scalar::from(*j) != k)
             .map(|j| Scalar::from(j as u32).expect_nonzero("index can not be zero"))
             .fold(Scalar::one(), |acc, j| {
@@ -335,26 +332,30 @@ impl<H: Digest<OutputSize = U32> + Clone, CH: Digest<OutputSize = U32> + Clone, 
         let binding_coeff = Scalar::from_hash(
             self.nonce_coeff_hash
                 .clone()
-                .add(agg_nonce_points)
+                .add(agg_nonce_points[0])
+                .add(agg_nonce_points[1])
                 .add(joint_key.joint_public_key)
                 .add(message),
         );
-        let group_commitment =
+
+        let (combined_agg_nonce, nonces_need_negation) =
             g!({ agg_nonce_points[0] } + binding_coeff * { agg_nonce_points[1] })
                 .normalize()
-                .expect_nonzero("impossibly unlikely, input is a hash");
+                .expect_nonzero("impossibly unlikely, input is a hash")
+                .into_point_with_even_y();
+
         let challenge = Scalar::from_hash(
             self.schnorr
                 .challenge_hash()
-                .add(binding_coeff.clone())
+                .add(combined_agg_nonce)
                 .add(joint_key.joint_public_key)
                 .add(message),
         );
 
         SignSession {
             binding_coeff,
-            agg_nonces: agg_nonce_points,
-            group_commitment,
+            nonces_need_negation,
+            agg_nonce: combined_agg_nonce,
             challenge,
             nonces: nonce_map,
         }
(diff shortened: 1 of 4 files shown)


sanket1729:
Thanks for looking at this. Looking forward to the PR.

- Would this be BIP 340 compatible? 

- Can this support BIP 32 derivations?

nickfarrow:
> * Would this be BIP 340 compatible?

We're making sure joint public keys have even Y coordinates, and are using FROST specific key-prefixing by tending to hash all public information where available. As well as tagged hashes. So i believe so!

> * Can this support BIP 32 derivations?

Currently this code produces a single joint public key for the multisignature (which can be tweaked). I'm actually not too sure how exactly address derivation will work. But my naive understanding would be that the joint public key could somehow be used for the BIP32 parent public key, mabye @LLFourn can chime in here

schnorr_fun/src/frost.rs
```diff
@@ -889,71 +889,55 @@ mod test {
...
-                    }
-                }
+            // use a boolean mask for which t participants are signers
+            let mut signer_mask = vec![true; threshold as usize];
+            signer_mask.append(&mut vec![false; (n_parties - threshold) as usize]);
+            // shuffle the mask for random signers (roughly shuffled and deterministic based on signers_mask_seed)
+            let mask_bytes = signers_mask_seed.to_be_bytes();
+            for i in 0..n_parties {
+                let rand_pos = mask_bytes[(i % mask_bytes.len() as u32) as usize] % n_parties as u8;
+                signer_mask.swap(i as usize, rand_pos as usize);
```
LLFourn, line 899: I think there is a shuffle function in `rand`.

nickfarrow:
doc test now seems to be failing because of some unrelated file
```
 Documenting ecdsa_fun v0.7.1 (/home/runner/work/secp256kfun/secp256kfun/ecdsa_fun)
warning: unresolved link to `derive_nonce`
   --> secp256kfun/src/nonce.rs:148:41
    |
148 | /// In general it's better to use the [`derive_nonce`] macro than to call
    |                                         ^^^^^^^^^^^^ no item named `derive_nonce` in scope
```

schnorr_fun/src/frost.rs
nickfarrow: ok cool, was just about to ask this. So OK to use rand in proptest? Using signers_mask_seed as seed for randomness

LLFourn: oh good point. It's ok to use something implementing [`Rng`](https://docs.rs/rand/0.8.5/rand/seq/trait.SliceRandom.html#tymethod.shuffle) to call shuffle as long as it's deterministic. I wonder if proptest has an easy way to get the rng it uses internally? See: https://docs.rs/proptest/latest/proptest/test_runner/struct.TestRunner.html#method.rng

nickfarrow: I couldnt figure out how to use that `.rng` method without getting rid of the convenient `proptest!` macro. Instead i've switched to:
```
signer_mask.shuffle(&mut TestRng::deterministic_rng(RngAlgorithm::ChaCha);
```
where `TestRng::deterministic_rng(RngAlgorithm::ChaCha)` seems to provide the necessary deterministic randomness

nickfarrow: Actually I don't think this is doing quite what we want, it doesn't vary as the proptest runs. Trying again

nickfarrow: I can't figure out a nice way to get the current run context's RNG without completely scrapping the macro and doing everything manually.

Leaning towards returning to using a proptest argument for the signer mask seed. I am thinking of using a rand integer that is then used as a seed for RNG.

See this branch & line: 
https://github.com/nickfarrow/secp256kfun/commit/7d956998080a9978021886d0634b311446ce0694#diff-65e82911697a71b2673ac046e6851aa3cfa3296ad1735eb1c283f381d131fe1fR898
Random 32 bytes being fed into ChaCha seed
image
(will force push here unless we can come up with a better idea!)

schnorr_fun/src/frost.rs
```diff
@@ -0,0 +1,1202 @@
...
+                    g!({ acc[0] } + { nonce.0[0] }),
+                    g!({ acc[1] } + { nonce.0[1] }),
+                ]
+            })
+            .iter()
+            .map(|agg_nonce| {
+                agg_nonce
+                    .normalize()
+                    .mark::<NonZero>()
+                    .expect("aggregate nonce should be non-zero")
```
LLFourn, line 713: it should not be but it can be :). I'd say we handle this case the same way as in `musig`.

LLFourn (changes requested):
Fantasitc work. API looks great. I've pointed out some internal issues that should be easy to fix.

schnorr_fun/src/frost.rs
```diff
@@ -0,0 +1,1166 @@
...
+
+    /// Create a scalar polynomial where the first coefficient is a specified secret and
+    /// the remaining coefficients are random.
+    ///
+    /// This will be used for creating secret polynomials with known & reproducable secrets.
+    pub fn random_using_secret(
+        n_coefficients: u32,
+        secret: Scalar,
+        rng: &mut (impl RngCore + CryptoRng),
+    ) -> Self {
```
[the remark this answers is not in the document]
nickfarrow: going to use `derive_nonce!` now, could make a single Nonce type (not binonce)

schnorr_fun/src/frost.rs
```diff
@@ -0,0 +1,1166 @@
...
+/// A joint FROST key
+#[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Deserialize, serde::Serialize),
+    serde(crate = "serde_crate")
+)]
+pub struct FrostKey {
+    /// The joint public key of the FROST multisignature.
+    pub joint_public_key: Point<Normal>,
```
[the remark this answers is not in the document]
nickfarrow: ahh, that explains the getter functions

schnorr_fun/src/frost.rs
```diff
@@ -0,0 +1,1166 @@
...
+        KeyGen: KeyGen,
+        my_index: u32,
+        secret_shares: Vec<Scalar<Secret, Zero>>,
+        proofs_of_possession: Vec<Signature>,
+    ) -> Result<(Scalar, FrostKey), FinishKeyGenError> {
+        assert_eq!(
+            secret_shares.len(),
+            KeyGen.frost_key.verification_shares.len()
+        );
+        assert_eq!(secret_shares.len(), proofs_of_possession.len());
```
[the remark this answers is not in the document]
nickfarrow: It makes the following zips and iterators messier, but could unpack at the begining of the function with
`let (secret_shares, proofs_of_possession) = shares_and_pops.iter().cloned().unzip();`

schnorr_fun/src/frost.rs
```diff
@@ -0,0 +1,1202 @@
...
+#[derive(Debug, Clone)]
+pub enum NewKeyGenError {
+    /// Received polynomial is of differing length.
+    PolyDifferentLength(usize),
+    /// Number of parties is less than the length of polynomials specifying the threshold.
+    NotEnoughParties,
+    /// Frost key is zero. This should be impossible, likely has been maliciously chosen.
+    ZeroFrostKey,
+    /// Verification share is zero. This should be impossible, likely has been maliciously chosen.
+    ZeroVerificationShare,
```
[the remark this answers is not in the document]
nickfarrow: Was thinking maybe you can only get a joint poly to evaluate to zero at index >0 if you negate all the other participant's polynomials, but yeah you'd fall under one of the other errors

So just `.expect` the `.mark::<NonZero>()` Result? or change to `Zero`?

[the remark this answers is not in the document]
LLFourn: Hmm I think leave it like it is. The correct solution is to allow it to be `Zero` but we can't serlaize `Zero` points yet. We can fix that but let's do it in another PR.

nickfarrow:
Paper of interest: Threshold Signatures with Private Accountability (no eprint yet but there is a [presentation](https://www.youtube.com/watch?v=TD8-QsGr0Zk) and [upcoming seminar (AEDT - 11 Oct 2022 11:00 AM)](https://www.monash.edu/it/ssc/cybersecurity/seminars/2022/threshold-signatures-with-private-accountability)

LLFourn (approved):
I rebased and made one small fixup. Let's merge this. Unfinished business:

1. Nonce generation API can use some refinement
2. verification shares should be allowed to be zero
3. nonce generation should be more like ietf spec. (probably more things should be -- we should read through it and see what's different).

PR #86, opened 2022-05-02
Only tweak with the new tweak contribution

nickfarrow:
`self.agg_key` already contains any previous `self.tweak` so we do not need to add `self.tweak + tweak`, just add the new `tweak`.

Tests for multiple tweak coming with FROST PR / MuSig rework PR.

There is probably a better way to do
```
let new_tweak = s!(0 + tweak).mark::<Public>();
```
which avoids 
 > the trait bound `impl ZeroChoice: DecideZero<_>` is not satisfied
the trait `DecideZero<_>` is not implemented for `impl ZeroChoice`

### DIFF
--- a/schnorr_fun/src/musig.rs
+++ b/schnorr_fun/src/musig.rs
@@ -209,11 +209,13 @@ impl KeyList {
     /// unusual case that the tweak is exactly equal to the negation of the aggregated secret key
     /// it returns `None`.
     pub fn tweak(&self, tweak: Scalar<impl Secrecy, impl ZeroChoice>) -> Option<Self> {
-        let mut tweak = s!(self.tweak + tweak).mark::<Public>();
-        let (agg_key, needs_negation) = g!(self.agg_key + tweak * G)
+        let new_tweak = s!(0 + tweak).mark::<Public>();
+        let (agg_key, needs_negation) = g!(self.agg_key + new_tweak * G)
             .mark::<NonZero>()?
             .into_point_with_even_y();
 
+        // Store accumulated tweak
+        let mut tweak = s!(self.tweak + tweak).mark::<Public>();
         tweak.conditional_negate(needs_negation);
 
         let needs_negation = self.needs_negation ^ needs_negation;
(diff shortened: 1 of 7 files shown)


schnorr_fun/src/musig.rs
```diff
@@ -209,11 +209,13 @@ impl KeyList {
     /// unusual case that the tweak is exactly equal to the negation of the aggregated secret key
     /// it returns `None`.
     pub fn tweak(&self, tweak: Scalar<impl Secrecy, impl ZeroChoice>) -> Option<Self> {
-        let mut tweak = s!(self.tweak + tweak).mark::<Public>();
-        let (agg_key, needs_negation) = g!(self.agg_key + tweak * G)
+        let new_tweak = s!(0 + tweak).mark::<Public>();
```
LLFourn, line 212: what are you trying to achieve here? clone?

nickfarrow: Hacky way of avoiding the `ZeroChoice` in `g!()` which needs `DecideZero` trait. What is the proper way?

LLFourn: What's the error you're getting?

nickfarrow: If tweak is ZeroChoice:
```
error[E0277]: the trait bound `impl ZeroChoice: DecideZero<_>` is not satisfied
   --> schnorr_fun/src/musig.rs:213:41
    |
213 |         let (agg_key, needs_negation) = g!(self.agg_key + tweak * G)
    |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `DecideZero<_>` is not implemented for `impl ZeroChoice`
    |
note: required by a bound in `scalar_mul_point`
   --> /home/nick/repos/secp256kfun/secp256kfun/src/op.rs:49:9
    |
49  |     Z1: DecideZero<Z2>,
    |         ^^^^^^^^^^^^^^ required by this bound in `scalar_mul_point`
```
I'm currently forcing new_tweak to not be zero choice with `s!(0 + tweak).mark::<Public>();`

PR #93, opened 2022-05-06
Make musig nonces usable in FROST

nickfarrow:
Upcoming FROST PR needs the ability to conditionally negate nonces

### DIFF
--- a/schnorr_fun/src/binonce.rs
+++ b/schnorr_fun/src/binonce.rs
@@ -1,0 +1,116 @@
+//! Binonces for Musig and FROST Signature Schemes
+//!
+//! A binonce is a pair of public points used in the MuSig and FROST signature schemes.
+//! Your public nonces are derived from scalars which must be kept secret.
+//! Derived binonces should be unique and and must not be reused for signing under any circumstances
+//! as this can leak your secret key.
+use secp256kfun::{g, marker::*, Point, Scalar, G};
+
+/// A nonce (pair of points) that each party must share with the others in the first stage of signing.
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub struct Nonce(pub [Point; 2]);
+
+impl Nonce {
+    /// Reads the pair of nonces from 66 bytes (two 33-byte serialized points).
+    pub fn from_bytes(bytes: [u8; 66]) -> Option<Self> {
+        let R1 = Point::from_slice(&bytes[..33])?;
+        let R2 = Point::from_slice(&bytes[33..])?;
+        Some(Nonce([R1, R2]))
+    }
+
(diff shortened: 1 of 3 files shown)


LLFourn (changes requested):
Let's move this to a small module called `binonce` that both MuSig and FROST can use.

schnorr_fun/src/binonce.rs
```diff
@@ -0,0 +1,170 @@
...
+//! Using a [`Synthetic`] nonce generator will mean you don't have to worry about passing a
+//! unique `sid` (session id) to this function for each signing session. The downside is that
+//! you must recall the result of `gen_nonces` somewhere and store it for use when you want to
+//! start the signing session.
+//!
+//! Using a [`Deterministic`] nonce generator means you **must** never start two signing
+//! sessions with nonces generated from the same `sid`. If you do your secret key will be
+//! recoverable from the two partial signatures you created with the same nonce. The upside is
+//! that you can rederive the nonces with the `sid` you originally passed
+//! to `gen_nonces` without having to store the output.
```
LLFourn, line 17: Hmm but don't we have `gen_nonce` on `MuSig` and `Frost`*? Isn't that where this documentation goes? I'd expect this module to be private and to just re-export the `Nonce` and `NonceKeyPair` types from both MuSig and Frost.

schnorr_fun/src/binonce.rs
```diff
@@ -0,0 +1,170 @@
...
+//!     nonce_gen => schnorr.nonce_gen(),
+//!     secret => keypair.secret_key(),
+//!     public => [ b"r2", keypair.public_key(), b"my-very-unique-sid"]
+//! );
+//! let R1 = g!(r1 * G).normalize();
+//! let R2 = g!(r2 * G).normalize();
+//! let keypair = NonceKeyPair {
+//!     public: Nonce([R1, R2]),
+//!     secret: [r1, r2],
+//! };
```
LLFourn, line 46: I don't think we want people constructing these manually.

schnorr_fun/src/lib.rs
```diff
@@ -22,6 +22,8 @@ extern crate serde_crate as serde;
 pub use secp256kfun as fun;
 pub use secp256kfun::nonce;
 
+/// Binonce for Musig and FROST
+pub mod binonce;
```
LLFourn, line 26: I think make this private and re-export stuff from it in both `musig` and `frost`. i.e. internally we are re-using it but people using the API can't tell that and always import the one from the API they are using.

LLFourn (changes requested):

LLFourn (approved):
LGTM

PR #94, opened 2022-05-06
MuSig rework - API design like FROST

nickfarrow:
Make MusiG API more like FROST implementation in https://github.com/LLFourn/secp256kfun/pull/84

* Removes `Party::Local` and `Party::Remote`, now and treat local/remote public things the same.
e.g.
```
let keylist = musig.new_keylist(vec![
    Party::Remote(p1_pubkey),
    Party::Local(my_keypair),
    Party::Remote(p3_pubkey),
]);
```
is now
```
let keylist = musig.new_keylist(vec![
    public_key1,
    public_key2,
    public_key3,
]);
```
* Nonces are now used:
```
let nonces = vec![p1_nonce.public, p2_nonce.public, p3_nonce.public];
```
as opposed to inputting local & remote nonces separately into functions.
* Handle nonce negation like with FROST: keylist tracks `needs_negation`
* Updates to latest signing tests from secp256k1-zkp (all the reference tests in the draft BIP excluding the tweak tests).

ALSO
* Removes deterministic signing functions
* Removes secret.clean(), sessions can now be reused but we need to make sure nonces are not. I got rid of mentions to clean()

This PR does not include tweak tests, and is not yet compatible with the BIP. The changes and tests for the ordinary tweaks will follow in a separate PR.


LLFourn (approved):
This LGTM only minor nits. I'll pass through again tomorrow.

LLFourn:
Nice work. I pushed a commit to fix up comments.

PR #95, opened 2022-05-15
Allow for ordinary (non xonly) tweaking of MuSig keys

nickfarrow:
`tweak(tweak, is_xonly)` now takes a bool as to whether the new tweak should update the accumulated needs negation. Also adds the remaining tests from the draft BIP.

* If the tweak is not x-only (an ordinary tweak), then never negate the aggregate key prior to applying the tweak, nor negate the existing tweak. Also do not update the MuSig keylist's `accum_needs_negation`. 

* If the tweak is x-only, then do the negation if required as usual.

I changed the name to `accum_needs_negation` so we can tell the code is using this spec. Not yet added to upcoming FROST PR.

I think we should review the updated logic together before merging.

see
https://github.com/jonasnick/bips/blob/musig2/bip-musig2.mediawiki#Tweaking_Definition
 https://github.com/jonasnick/bips/issues/13

### DIFF
--- a/schnorr_fun/src/musig.rs
+++ b/schnorr_fun/src/musig.rs
@@ -121,11 +121,11 @@ pub struct KeyList {
     /// The coefficients of each key
     coefs: Vec<Scalar<Public>>,
     /// The aggregate key
-    agg_key: Point<EvenY>,
+    agg_key: Point<Normal>,
     /// The tweak on the aggregate key
     tweak: Scalar<Public, Zero>,
     /// Whether this aggregate key needs negation.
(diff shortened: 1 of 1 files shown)


nickfarrow:
ah I had the title right the first time. Ordinary tweaks are for BIP32, `is_xonly=false`,

LLFourn (approved):
Thanks @nickfarrow. I think we can merge this now and try and improve upon it.

I think the only thing that isn't compatible with spec is the fallback to `G` when aggregate `R` value is zero right?

PR #96, opened 2022-06-13
[musig] keypairs never need negation

LLFourn:
Since we do the negation check when we create them. We can assume this if take a `KeyPair` when signing instead. I think this is more in the spirit of the library. I might be about to change this back but I wanted to commit in master with it like this in there.


nickfarrow:
makes sense, LGTM!

LLFourn:
superseded by #100

PR #97, opened 2022-06-14
[musig2] two phase key aggregation

LLFourn:
Users must transition from a `AggKey` into an `Bip340AggKey` before starting a signing session. This is the idea:

1. When it is a `AggKey` you can only apply ordinary tweaks (i.e. bip32)
2. When it is a `Bip340Aggkey` you can only apply "xonly" tweaks. (i.e. taproot)

The code and spec makes way more sense to me now. We are down to a single `needs_negation` boolean we keep around. We can't implement all the spec tests for this since one test requires a ordinary tweak after a "xonly" tweak. I think I'll suggest removing that from the spec.

### DIFF
--- a/schnorr_fun/src/musig.rs
+++ b/schnorr_fun/src/musig.rs
@@ -34,9 +34,9 @@
 //! // start the signing session
 //! let session = musig.start_sign_session(&keylist, nonces, message).unwrap();
 //! // sign with our single local keypair
-//! let my_sig = musig.sign(&keylist, 0, my_keypair.secret_key(), my_nonce, &session);
-//! # let p2_sig = musig.sign(&keylist, 1, kp2.secret_key(), p2_nonce, &session);
-//! # let p3_sig = musig.sign(&keylist, 2, kp3.secret_key(), p3_nonce, &session);
+//! let my_sig = musig.sign(&keylist, 0, &my_keypair, my_nonce, &session);
+//! # let p2_sig = musig.sign(&keylist, 1, &kp2, p2_nonce, &session);
+//! # let p3_sig = musig.sign(&keylist, 2, &kp3, p3_nonce, &session);
 //! // receive p2_sig and p3_sig from somewhere and check they're valid
 //! assert!(musig.verify_partial_signature(&keylist, &session, 1, p2_sig));
 //! assert!(musig.verify_partial_signature(&keylist, &session, 2, p3_sig));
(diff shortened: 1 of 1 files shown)


schnorr_fun/src/musig.rs
```diff
@@ -124,78 +125,112 @@ pub struct KeyList {
...
 
-    /// Tweak the aggregate MuSig public key with a scalar so that the resulting key is equal to the
-    /// existing key plus `tweak * G`. The tweak mutates the public key while still allowing
-    /// the original set of signers to sign under the new key.
+    /// Add a scalar `tweak` to aggregate MuSig public key.
     ///
-    /// This is how you embed a taproot commitment into a key.
+    /// The resulting key is equal to the existing key plus `tweak * G`. The tweak mutates the
+    /// public key while still allowing the original set of signers to sign under the new key.
+    /// This function is appropriate for doing [BIP32] tweaks before calling `into_bip340_key`.
```
nickfarrow, line 145: Thinking maybe move these two lines on top of the ones above? Just to be sure people see it and don't use the wrong tweak

schnorr_fun/src/musig.rs
```diff
@@ -124,78 +125,112 @@ pub struct KeyList {
...
     tweak: Scalar<Public, Zero>,
-    /// Whether this aggregate key needs negation.
-    accum_needs_negation: bool,
 }
 
-impl KeyList {
-    /// The `XOnly` aggregated key for the keylist.
-    pub fn agg_public_key(&self) -> XOnly {
-        self.agg_key.to_xonly()
+impl AggKey {
```
nickfarrow, line 130: Could rename `AggKey` to `MuSigKey` like we are planning with `FrostKey` which I quite like. But maybe this makes `Bip340MuSigKey` too long

With frost we will also have a `Bip340FrostKey` / `Bip340AggKey`

schnorr_fun/src/musig.rs
```diff
@@ -124,78 +125,112 @@ pub struct KeyList {
...
+
+    /// An iterator over the **public keys** of each party in the aggregate key.
     pub fn keys(&self) -> impl Iterator<Item = XOnly> + '_ {
         self.parties.iter().map(|xonly| *xonly)
     }
 
-    /// Tweak the aggregate MuSig public key with a scalar so that the resulting key is equal to the
-    /// existing key plus `tweak * G`. The tweak mutates the public key while still allowing
-    /// the original set of signers to sign under the new key.
+    /// Add a scalar `tweak` to aggregate MuSig public key.
```
nickfarrow, line 141: Both tweaks functions take scalar tweaks as inputs so perhaps this isn't the best description. But I also do not like `ordinary` tweaks..

schnorr_fun/src/musig.rs
LLFourn: Yeah let's apply the same thing to FROST and see what naming we like best. I think what you've suggested is right.

LLFourn:
superseded by #37

PR #100, opened 2022-07-01
[musig2] use plain points as inputs

LLFourn:
Experimenting with the idea I put forward here: https://github.com/jonasnick/bips/issues/32. 

 Since this PR builds upon #97 it already has two phase key aggregation so the change is very simple. The other thing to note is that we were already "pre-negating" the secret keys of keypairs so we don't get any reduction in the number of negation flags we have to keep around (we were down to one already).


LLFourn:
I've just pushed large changes making this compatible with and passing the tests from https://github.com/jonasnick/bips/pull/37

PR #103, opened 2022-07-05
alt: Bump secp256k1 version to 0.22

h4sh3d:
Based on #99

Alternative solution to #99 with new type `Bip340NoAux` for tests only. I created another PR to keep the first implementation visible for history, I let you close #99 when this get merged.

Do you want to keep the proptest generated failed test case seed? Or should I amend the commit and remove it?


LLFourn:
Thanks!

PR #108, opened 2022-08-23
ci: add armv7 target

delta1:
Adds the `armv7-unknown-linux-gnueabihf` target to CI strategy matrix, but fails to compile `secp256kfun_parity_backend` in cargo test, even though cargo build works?

`cargo build --target armv7-unknown-linux-gnueabihf -p secp256kfun_parity_backend` -> works

`cargo test --target armv7-unknown-linux-gnueabihf -p secp256kfun_parity_backend` -> fails

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         args: [
             "--no-default-features",
@@ -34,6 +35,7 @@ jobs:
             ""
         ]
         rust: [nightly, stable]
+        target: ["x86_64-unknown-linux-gnu", "armv7-unknown-linux-gnueabihf"]
         exclude:
           - rust: stable
             args: --all-features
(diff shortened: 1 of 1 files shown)


nickfarrow:
Does adding this to `.cargo/config` help?
```
# You need to install the "arm-none-linux-gnueabihf" linker on your system
[target.armv7-unknown-linux-gnueabihf]
linker = "arm-none-linux-gnueabihf-gcc"
```

It gets me to a different error
```
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running unittests src/lib.rs (/home/nick/repos/secp256kfun/target/armv7-unknown-linux-gnueabihf/debug/deps/secp256kfun_parity_backend-94ce32d7b35c0a66)
error: test failed, to rerun pass '-p secp256kfun_parity_backend --lib'

Caused by:
  could not execute process `/home/nick/repos/secp256kfun/target/armv7-unknown-linux-gnueabihf/debug/deps/secp256kfun_parity_backend-94ce32d7b35c0a66` (never executed)
```
Or is that just because i'm on the wrong architecture?

Whereas on my RPi I get `running 0 tests`.

.github/workflows/test.yml
```diff
@@ -45,9 +47,24 @@ jobs:
         with:
           profile: minimal
           toolchain: ${{ matrix.rust }}
+          target: ${{ matrix.target }}
           override: true
+      - name: Install cargo cross
+        run: |
+          curl -L https://github.com/cross-rs/cross/releases/download/v0.2.4/cross-x86_64-unknown-linux-gnu.tar.gz | tar xvz
+          sudo mv cross cross-util /usr/bin
+          which cross
```
LLFourn, line 56: In https://github.com/actions-rs/cargo#cross-compilation that you linked it says that it will install `cross` automatically. Did that not work here?

.github/workflows/test.yml
```diff
@@ -45,9 +47,24 @@ jobs:
...
+      - name: Install compiler for armhf arch
+        if: matrix.target == 'armv7-unknown-linux-gnueabihf'
+        run: |
+          sudo apt-get update
+          sudo apt-get install gcc-arm-linux-gnueabihf
```
LLFourn, line 61: isn't the point of `cross` to not have to do this?

.github/workflows/test.yml
delta1: it does work but uses `cargo install cross` I think which just takes longer :man_shrugging: will change it though

.github/workflows/test.yml
delta1: that's a good point, will try without it!

delta1:
Updated, but seems there may have been a breaking change in `cross` 

You can see the failures here: https://github.com/delta1/secp256kfun/runs/8186507180?check_suite_focus=true

delta1:
> I think their cache was restored from the wrong target

Updated the rust-cache action to v2

LLFourn:
LGTM. Thanks @delta1!

PR #119, opened 2022-09-15
FROST cfrg spec

nickfarrow:
Also create keygen hash like Jesse's https://github.com/ElementsProject/secp256k1-zkp/blob/6c1cf4e0caac3a933b4f8cdbe14be05ea4c7c0d9/src/modules/frost/keygen_impl.h#L210

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -58,7 +58,8 @@
 //! #        proofs_of_possession.clone(),
 //! #    )
 //! #    .unwrap();
-//! // signing parties must use a common set of nonces when creating signature shares
+//! // signing parties must use a common set of nonces when creating signature shares.
+//! // nonces can be derived from a session id that use includes publicly known values.
 //! let verification_shares_bytes: Vec<_> = frost_key
 //!     .verification_shares()
 //!     .map(|share| share.to_bytes())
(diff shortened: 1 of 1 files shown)


nickfarrow:
The latest [draft-irtf-cfrg-frost-08](https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-08.html) differs from v[04](https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-08.html) by having a per signer rho instead of a single group rho. See https://github.com/cfrg/draft-irtf-cfrg-frost/issues/214. Thinking I'll match 04 first, then explore per signer rhos.

Current discrepancies: 
* My commitment hash participant index uses u8 (L862) they uses u16
* Currently using the same binding coeff hash on the message

In Keygen, Jesse's implementation uses some compute_indexhash so `indexhash = tagged_hash(pk || idx) `
https://github.com/ElementsProject/secp256k1-zkp/blob/6c1cf4e0caac3a933b4f8cdbe14be05ea4c7c0d9/src/modules/frost/keygen_impl.h#L56

nickfarrow:
Currently we use BTreeMaps for sorting nonces:
```
        let mut nonce_map: BTreeMap<_, _> =
            nonces.into_iter().map(|(i, nonce)| (i, nonce)).collect();
```

PR #123, opened 2022-10-03
Frost reuse message in test and some docs grammar

nickfarrow:
Your commits covered the other changes I had, or made them obsolete 👍

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -620,7 +620,7 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
     ///
     /// ## Return value
     ///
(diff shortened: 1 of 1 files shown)


schnorr_fun/src/frost.rs
```diff
@@ -620,7 +620,7 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
     ///
     /// ## Return value
     ///
-    /// Returns `bool` true if the proof of possession matches the point
+    /// Returns `bool` true if the proof of possession matches the point polynomial
```
LLFourn, line 623:
```suggestion
    /// Returns `bool` true if the proof of possession matches the point
```

The PoP is for the point only.

LLFourn (approved):
Thanks.

PR #127, opened 2022-10-06
Remove ScalarPoly and PointPoly

LLFourn:
II'm still thinking about this a bit.

What if we instead just make a `ScalarVec` and `PointVec` types that had `poly_eval` function which treats them as the coefficients of a polynomial.

Ok better yet why not just make extension traits for `Vec<Point>` and `Vec<Scalar>`?

@nickfarrow thoughts?

## UPDATE

I decided to just trash both these types and provide helper methods in the `frost` module. This is simpler. Maybe one day we want to provide some tools for dealing with vectors/polynomials of points/scalars but not today!


schnorr_fun/src/frost.rs
```diff
@@ -1,112 +1,77 @@
...
-//! Signatures]*.
-//!
-//! > ⚠ At this stage this implementation is for API exploration purposes only. The way it is
-//! currently implemented is not proven secure.
-//!
-//! [FROST: Flexible Round-Optimized Schnorr Threshold Signatures]: <https://eprint.iacr.org/2020/852.pdf>
-//! [secp256k1-zkp]: <https://github.com/ElementsProject/secp256k1-zkp/pull/138>
-//! [Security of Multi- and Threshold Signatures]: <https://eprint.iacr.org/2021/1375.pdf>
-//! [`musig`]: crate::musig
+//! The FORST threshold multisignature scheme.
```
nickfarrow, line 1: FORST -> FROST

PR #128, opened 2022-10-13
Frost and MuSig joint key Point type parameter

nickfarrow:
* Change musig and frost keys to allow for any point type
* keygen creates Normal points
* signing operations are done with EvenY points

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -5,10 +5,9 @@
 //! ```
 //! use schnorr_fun::{
 //!     frost,
-//!     Schnorr,
 //!     Message,
-//!     fun::Scalar
 //! };
+//! use rand_chacha::ChaCha20Rng;
 //! use sha2::Sha256;
 //! // use sha256 to produce deterministic nonces -- be careful!
 //! let proto = frost::new_with_deterministic_nonces::<Sha256>();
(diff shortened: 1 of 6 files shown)


schnorr_fun/src/frost.rs
```diff
@@ -473,30 +438,50 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
-            session_id,
-            public_key,
-            message,
-        )
+    ) -> R
+    where
+        <R as SeedableRng>::Seed: From<GenericArray<u8, U32>>,
+    {
+        let message = message.unwrap_or(Message::raw(b""));
+        let msg_len = (message.len() as u64).to_be_bytes();
```
nickfarrow, line 467: These could be moved into the macro?

schnorr_fun/tests/frost_prop.rs
```diff
@@ -54,13 +56,18 @@ proptest! {
         let sid = b"frost-prop-test".as_slice();
         let message = Message::plain("test", b"test");
 
-        let nonces: Vec<_> = signer_indexes.iter().map(|i|
-            proto.gen_nonce(
-                &secret_shares[*i],
+        let mut nonce_rngs: Vec<ChaCha20Rng> = secret_shares.iter().map(|secret_share| {
```
nickfarrow, line 59: `RngAlgorithm::ChaCha` is not `Seedable` so using `ChaCha20Rng`... Doesn't seem ideal

LLFourn:
Sorry I somehow goofed the thing when merged the parent PR and deleted the branch. Can you reopen?

schnorr_fun/tests/frost_prop.rs
LLFourn: `ChaCha20Rng` is the right choice.

schnorr_fun/src/frost.rs
LLFourn: it could be moved into `HashAdd`.

PR #129, opened 2022-10-14
Remove .borrow from derive_nonce_rng

LLFourn:
It breaks type inference and is unnecessary.


nickfarrow:
This fixes my type inference error, LGTM!

PR #131, opened 2022-10-17
Frost and MuSig joint key Point type parameter

nickfarrow:
Recreating https://github.com/LLFourn/secp256kfun/pull/128 minus the cherry-picked commit from https://github.com/LLFourn/secp256kfun/pull/129/
* Change musig and frost keys to allow for any point type
* keygen creates Normal points
* signing operations are done with EvenY points

### DIFF
--- a/schnorr_fun/Cargo.toml
+++ b/schnorr_fun/Cargo.toml
@@ -17,6 +17,7 @@ keywords = ["bitcoin", "schnorr"]
 features = ["all"]
 
 [dependencies]
+rand = "0.8"
 secp256kfun = { path = "../secp256kfun", version = "0.7.1",  default-features = false }
 serde_crate = { package = "serde", version = "1.0", default-features = false, optional = true, features = ["derive", "alloc"] }
 
@@ -44,7 +45,7 @@ harness = false
 
 [features]
 default = ["std"]
-all = ["std","serde", "libsecp_compat", "proptest"]
+all = ["std", "serde", "libsecp_compat", "proptest"]
 alloc = ["secp256kfun/alloc"]
 std = ["alloc", "secp256kfun/std"]
 serde = ["serde_crate", "secp256kfun/serde"]
--- a/schnorr_fun/Cargo.toml
+++ b/schnorr_fun/Cargo.toml
@@ -17,7 +17,6 @@ keywords = ["bitcoin", "schnorr"]
 features = ["all"]
 
 [dependencies]
-rand = "0.8"
 secp256kfun = { path = "../secp256kfun", version = "0.7.1",  default-features = false }
 serde_crate = { package = "serde", version = "1.0", default-features = false, optional = true, features = ["derive", "alloc"] }
 
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -4,6 +4,7 @@
 //!
 //! ```
 //! use schnorr_fun::{
+//!     binonce::NonceKeyPair,
 //!     frost,
 //!     Message,
 //! };
@@ -59,9 +60,9 @@
 //! let message =  Message::plain("my-app", b"chancellor on brink of second bailout for banks");
 //! // Generate nonces for this signing session.
 //! let session_id = b"signing-cool-message-attempt-1".as_slice(); // ⚠ must be different for every session
-//! let mut nonce_rng: ChaCha20Rng = proto.gen_nonce_rng(&my_secret_share, session_id, Some(frost_key.public_key()), Some(message));
+//! let mut nonce_rng: ChaCha20Rng = proto.gen_nonce_rng(&frost_key, &my_secret_share, session_id, Some(message));
 //! let my_nonce = proto.gen_nonce(&mut nonce_rng);
-//! # let nonce3 = proto.gen_nonce(&mut nonce_rng);
+//! # let nonce3 = NonceKeyPair::random(&mut rand::thread_rng());
 //! // share your public nonce with the other signing participant(s)
 //! # let received_nonce3 = nonce3.public();
 //! // receive public nonces from other signers
@@ -146,18 +147,14 @@
 #![cfg(feature = "serde")]
 pub use crate::binonce::{Nonce, NonceKeyPair};
 use crate::{Message, Schnorr, Signature, Vec};
-use rand::SeedableRng;
 use secp256kfun::{
     derive_nonce_rng,
-    digest::{
-        generic_array::{typenum::U32, GenericArray},
-        Digest,
-    },
+    digest::{generic_array::typenum::U32, Digest},
     g,
     hash::{HashAdd, Tagged},
     marker::*,
     nonce::{self, AddTag, NonceGen},
-    rand_core::RngCore,
+    rand_core::{RngCore, SeedableRng},
     s, Point, Scalar, G,
 };
 use std::collections::BTreeMap;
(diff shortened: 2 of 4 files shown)


schnorr_fun/Cargo.toml
```diff
@@ -44,7 +45,7 @@ harness = false
 
 [features]
 default = ["std"]
-all = ["std","serde", "libsecp_compat", "proptest"]
+all = ["std", "serde", "libsecp_compat", "proptest"]
```
LLFourn, line 48: removed whitesspace

schnorr_fun/Cargo.toml
```diff
@@ -17,6 +17,7 @@ keywords = ["bitcoin", "schnorr"]
 features = ["all"]
 
 [dependencies]
+rand = "0.8"
```
LLFourn, line 20: We don't need this dependency on rand.

schnorr_fun/src/frost.rs
```diff
@@ -143,10 +143,16 @@
 //! ```
 //!
 //! Note that if a key generation sesssion fails you must always start a fresh session with a different session id.
+#![cfg(feature = "serde")]
 pub use crate::binonce::{Nonce, NonceKeyPair};
 use crate::{Message, Schnorr, Signature, Vec};
+use rand::SeedableRng;
```
LLFourn, line 149: just get this from `rand_core`.

schnorr_fun/src/frost.rs
```diff
@@ -473,30 +438,50 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
     /// Otherwise you can use your secret share of the frost key.
     ///
     /// The application should decide upon a unique `session_id`. If the `NonceGen` of this FROST
     /// instance is `Deterministic` then the `session_id` **must** be unique per signing session.
-    pub fn gen_nonce(
+    pub fn gen_nonce_rng<R: SeedableRng>(
         &self,
         secret: &Scalar,
         session_id: &[u8],
         public_key: Option<Point<impl Normalized>>,
```
LLFourn, line 460: the main point of this PR from my perspective was so you can pass in `FrostKey<T>` here and get the verification shares out and hash them too. Actually I think strictly speaking if you hash the threshold and all the verification shares that strictly determines the public key anyway (via lagrange interpolatoin).

I think last time we spoke I mentioned that we should just use `ChaCha20Rng`, add a feature flag here and make it concrete. After looking at it I changed my mind. I like what you've done here better. I think you can simplify the type constraint here though. Should just be `R: SeedableRng<Seed=[u8;32]>`

schnorr_fun/src/frost.rs
```diff
@@ -473,30 +438,50 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
+            // regardless of whether you pass in a Normal or EvenY point.
+            .map(|public_key| {
+                let public_key: Point<Normal> = public_key.normalize();
+                public_key.to_bytes()
+            })
+            .unwrap_or([0u8; 33]);
+        let rng: R = derive_nonce_rng!(
+            nonce_gen => self.schnorr.nonce_gen(),
+            secret => &secret,
+            public => [b"r1", pk_bytes, msg_len, message, sid_len, session_id],
```
LLFourn, line 481: "r1" doesn't make sense here and potentially harms the security. We have to guarantee that it is impossible to generate the same inputs as normal single signer `Schnorr`. I think this is guaranteed if we make the `pk_bytes` the first thing in.

Actually it would be ideal if we did this properly and simply cloned the nonce gen out of the `self.schnorr` and tagged it with `frost/nonce` and stored it in `Frost`. It already has the `NonceGen` type as a parameter so no real friction here.

schnorr_fun/src/frost.rs
```diff
@@ -206,13 +212,13 @@ impl<H: Tagged, NG> Frost<H, NG> {
 ///
 /// [`Frost::new_keygen`]
 #[derive(Clone, Debug)]
-pub struct KeyGen {
+pub struct KeyGen<T> {
     point_polys: Vec<Vec<Point>>,
     keygen_id: [u8; 32],
-    frost_key: FrostKey,
+    frost_key: FrostKey<T>,
```
LLFourn, line 218: This should just be `Normal`.

schnorr_fun/Cargo.toml
nickfarrow: I'm adding whitespace here, i think it was an rustfmt

schnorr_fun/src/frost.rs
```diff
@@ -276,26 +282,32 @@ impl core::fmt::Display for FinishKeyGenError {
 impl std::error::Error for FinishKeyGenError {}
 
 /// A joint FROST key
-#[derive(Clone, Debug, PartialEq)]
```
LLFourn, line 279: This should be `PartialEq`. It's quite unfortunate that we can't derive it here.... I would make an exception to the rule about not putting type constraints on struct type params and put `FrostKey<T: PointType>` here.

schnorr_fun/src/frost.rs
```diff
@@ -642,13 +631,13 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
     /// # Return value
     ///
     /// Your secret share and the [`FrostKey`]
-    pub fn finish_keygen(
+    pub fn finish_keygen<T>(
```
LLFourn, line 634: should be no type parameter here.

schnorr_fun/src/frost.rs
```diff
@@ -556,7 +541,7 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
     /// ## Return value
     ///
     /// Returns `bool` true if the proof-of-possession matches the point
-    fn verify_pop(&self, keygen: &KeyGen, point: Point, pop: Signature) -> bool {
+    fn verify_pop<T>(&self, keygen: &KeyGen<T>, point: Point, pop: Signature) -> bool {
```
LLFourn, line 544: should be no type parameter here.

schnorr_fun/src/frost.rs
```diff
@@ -60,8 +59,9 @@
 //! let message =  Message::plain("my-app", b"chancellor on brink of second bailout for banks");
 //! // Generate nonces for this signing session.
 //! let session_id = b"signing-cool-message-attempt-1".as_slice(); // ⚠ must be different for every session
-//! let my_nonce = proto.gen_nonce(&my_secret_share, session_id, Some(frost_key.public_key()), Some(message));
-//! # let nonce3 = proto.gen_nonce(&secret_share3, session_id, Some(frost_key.public_key()), Some(message));
+//! let mut nonce_rng: ChaCha20Rng = proto.gen_nonce_rng(&my_secret_share, session_id, Some(frost_key.public_key()), Some(message));
+//! let my_nonce = proto.gen_nonce(&mut nonce_rng);
+//! # let nonce3 = proto.gen_nonce(&mut nonce_rng);
```
LLFourn, line 64: probably a bad idea to get nonce randomness from another notional party even in hidden doc example. You can just let `nonce3 = binonce::NonceKeyPair::random(&mut rand::thread_rng())`.

LLFourn (changes requested):
Merging the two states of the key into one looks good. 

1. The type parameter should only be set to `T` in one function: `gen_nonce_rng`. `FrostKey` needs to be passed in here.
2. Delete the nonce generation stuff from `binonce` as the rng sutff supersedes it.
3. Add a `gen_nonce_rng` to the `MuSig` type which takes the `AggKey` in the same way you did it in `frost`.

nickfarrow:
I believe this PR is ready for merge. I think we will leave cloning `nonce_gen` into `Frost` in a later PR.

schnorr_fun/src/frost.rs
```diff
@@ -473,30 +435,51 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
     /// The application should decide upon a unique `session_id`. If the `NonceGen` of this FROST
     /// instance is `Deterministic` then the `session_id` **must** be unique per signing session.
-    pub fn gen_nonce(
+    pub fn gen_nonce_rng<T: PointType, R: SeedableRng<Seed = [u8; 32]>>(
         &self,
+        frost_key: &FrostKey<T>,
         secret: &Scalar,
         session_id: &[u8],
-        public_key: Option<Point<impl Normalized>>,
+        // public_key: Option<Point<impl Normalized>>,
```
LLFourn, line 458: I guess just remove this?

LLFourn (approved):
Thanks!

PR #132, opened 2022-10-19
Clean up feature flags make things compile with alloc and --no-default-features

LLFourn:
I've noticed that `secp256kfun` doesn't actually compile with `--no-default-features`. This is weird because we test it. The problem is that it works at the workspace root but if you cd into the directory and run the same command it breaks. According to https://github.com/rust-lang/cargo/issues/4753 this issue as been fixed ages ago but it's happening for me.

So this PR is for investigating this and cleaning up feature flags in general.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.60.0
           target: ${{ matrix.target }}
           override: true
       - uses: Swatinem/rust-cache@v2.0.0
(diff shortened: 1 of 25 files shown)


LLFourn:
I've pushed some commits to use the new cargo features available in [`v1.60.0`](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-160-2022-04-07). Now the dependencies section for `schnorr_fun` looks like:

```toml
[dependencies]
secp256kfun = { path = "../secp256kfun", version = "0.7.1",  default-features = false }
```

and that's it! I've managed to re-export `serde` which I could never figure out how to do before.

Now `ecdsa_fun`'s adaptor feature can be used without serde (this was impossible to arrange before).

I think I'll go ahead and try and add some "only if this feature is enabled" doc thingos and that will be it for this PR.

PR #140, opened 2022-12-15
[forst][musig] Clean up nonce generation

LLFourn:
Simplify nonce generation. We only provide a method where the user can supply a session id, secret, joint public key and we give them an RNG they can derive nonces from.

I removed optional arguments to force the user to focus on generating a session id that gives security guarantees and not hoping that combinations of message and public keys etc makes it unique.

Also `Frost` and `MuSig` now have own uniquely tagged nonce generator so there can be full domain separation between multisigs and single sigs with the same `Schnorr` instance.

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -61,7 +61,7 @@
 //! // Generate nonces for this signing session.
 //! // ⚠ session_id must be different for every signing attempt
 //! let session_id = b"signing-ominous-message-about-banks-attempt-1".as_slice();
-//! let mut nonce_rng: ChaCha20Rng = frost.gen_nonce_rng(&frost_key, &my_secret_share, session_id);
+//! let mut nonce_rng: ChaCha20Rng = frost.seed_nonce_rng(&frost_key, &my_secret_share, session_id);
 //! let my_nonce = frost.gen_nonce(&mut nonce_rng);
 //! # let nonce3 = NonceKeyPair::random(&mut rand::thread_rng());
 //! // share your public nonce with the other signing participant(s)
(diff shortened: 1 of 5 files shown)


schnorr_fun/src/frost.rs
```diff
@@ -440,50 +462,39 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
-    /// use the static first coefficient of your polynomial.
-    /// Otherwise you can use your secret share of the frost key.
+    /// The application should decide upon a unique `session_id` per call to this function. If the
+    /// `NonceGen` of this FROST instance is `Deterministic` then the `session_id` **must** be
+    /// unique per signing attempt -- even if the signing attempt fails to produce a signature you
+    /// must not reuse the session id, the resulting rng or anything derived from that rng again.
     ///
-    /// The application should decide upon a unique `session_id`. If the `NonceGen` of this FROST
-    /// instance is `Deterministic` then the `session_id` **must** be unique per signing session.
+    /// 💡 Before using this function write a short justification as to why you beleive you session id
```
nickfarrow, line 481: your* session id.

```
Note **it must be unique even across signing attempts for [same or] different messages**.
```

schnorr_fun/src/frost.rs
```diff
@@ -3,17 +3,17 @@
...
 //! use schnorr_fun::{
-//!     binonce::NonceKeyPair,
 //!     frost,
 //!     Message,
 //! };
 //! use rand_chacha::ChaCha20Rng;
 //! use sha2::Sha256;
 //! // use sha256 to produce deterministic nonces -- be careful!
-//! let proto = frost::new_with_deterministic_nonces::<Sha256>();
+//! let frost = frost::new_with_deterministic_nonces::<Sha256>();
```
nickfarrow, line 14: nice I prefer this as `frost` over `proto`

nickfarrow:
From an API perspective I think this is much clearer for the user,  that "session_id must be different for every signing attempt"/signing session

Am i understanding correctly that for a deterministic nonce gen:
```
let session_id = b"signing-ominous-message-about-banks-attempt-1".as_slice();
let mut nonce_rng: ChaCha20Rng = frost.gen_nonce_rng(&frost_key, &my_secret_share, session_id);
let my_nonce = frost.gen_nonce(&mut nonce_rng);
```
we would only ever use `nonce_rng` once? But with a synthetic noncegen, we can safely reuse `nonce_rng` to create more nonces in following sessions?

schnorr_fun/src/frost.rs
LLFourn: I had to change it for some reason but it worked when I changed it back!

LLFourn:
> we would only ever use `nonce_rng` once? But with a synthetic noncegen, we can safely reuse `nonce_rng` to create more nonces in following sessions?

No you can use the nonce rng to create as many nonces as you want for the session. You're right we need to explicitly state this and give an example. A bitcoin transaction is a good example.

Issue #31, opened 2020-10-07
Acronym is in all caps


tcharding:
The acronym `ECDSA` is in all caps in this repository. This is counter to Rust coding conventions so for users of the library it seems unusual. Is there a reason for this break in convention?

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn. In snake_case, acronyms and contractions are lower-cased: is_xid_start.

ref: https://rust-lang.github.io/api-guidelines/naming.html

LLFourn:
No there is no reason in particular. Though if we're going to be strict about this I guess it's time to enable clippy and find something that picks up these kinds of things.

In general while doing crypto implementations I violate the shit out of variable naming conventions e.g. `let X = g!(x * G)`. which is just necessary for clarity. I don't mind sticking to it for type names.

Let's do clippy and see if we can address that then: https://github.com/LLFourn/secp256kfun/issues/32

tcharding:
> In general while doing crypto implementations I violate the shit out of variable naming conventions e.g. let X = g!(x * G). which is just necessary for clarity. 

Yeah it almost broke my eyes when I started reading crypto code with things like `X_a` but I see now why its necessary and ones eyes get used to it.

Cheers

PR #147, opened 2023-02-07
Add clippy and resolve suggestions

nickfarrow:
Small clippy fixes and also removed the pointless `sid_len` from the `derive_nonce_rng!` that doesn't add any collision resistance. Though this will **break consistency of deterministic nonce rngs between releases**.

Prior to this fixup, the `&&`s have just been getting fully dereferenced right?
Like
```
Scalar::from_hash(hash.add(R).add(X).add(&&Message))
```

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -263,7 +263,7 @@ impl core::fmt::Display for NewKeyGenError {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
         use NewKeyGenError::*;
         match self {
-            PolyDifferentLength(i) => write!(f, "polynomial commitment from party at index {} was a different length", i),
+            PolyDifferentLength(i) => write!(f, "polynomial commitment from party at index {i} was a different length"),
             NotEnoughParties => write!(f, "the number of parties was less than the threshold"),
             ZeroFrostKey => write!(f, "The frost public key was zero. Computationally unreachable, one party is acting maliciously."),
         }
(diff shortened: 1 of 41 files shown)


schnorr_fun/src/frost.rs
```diff
@@ -493,14 +491,13 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
     ) -> R {
-        let sid_len = (session_id.len() as u64).to_be_bytes();
         let threshold_bytes = (frost_key.threshold() as u64).to_be_bytes();
         let pk_bytes = frost_key.public_key().to_xonly_bytes();
 
         let rng: R = derive_nonce_rng!(
             nonce_gen => self.nonce_gen(),
             secret => &secret,
-            public => [pk_bytes, threshold_bytes, sid_len, session_id],
+            public => [pk_bytes, threshold_bytes, session_id],
```
LLFourn, line 500: I'd prefer to keep this. Anything that is variable length should be length prefixed. I actually want to do this explicitly in the hashInto implementation.

LLFourn:
Cool. Thanks. can you add clippy to CI so we can finally close #32.

nickfarrow:
I noticed only a few uppercase acronyms (https://github.com/LLFourn/secp256kfun/issues/31)
```
type DLEQ = Eq<$mod::DLG<$len>, $mod::DL<$len>>;
```
Currently i have left them as clippy allowed

schnorr_fun/src/message.rs
```diff
@@ -37,6 +37,11 @@ impl<'a, 'b, S: Secrecy> Message<'a, S> {
         }
     }
 
+    /// Check if the message is empty with zero length
+    pub fn is_empty(&self) -> bool {
```
LLFourn, line 41: Did clippy make you do this?

LLFourn:
Thanks so much for this!

schnorr_fun/src/message.rs
nickfarrow: Yeah, something about implementing len but not empty

PR #148, opened 2023-02-13
Allow creating an n-of-n FROST key from a set of core shares

LLFourn:
This intended to demonstrate how we could re-imagine FROST key generation.
This allows generating an n-of-n FROST key from a list of n public keys. This is insecure without proofs of possession for each of the n keys . The intention is that from this n-of-n FROST key you could add new parties to make it a t-of-n by using an enrollment protocol see [1]. We don't have a proof of security with FROST signing with respect to generating an n-of-n key like this let alone later converting it to a t-of-n key.

[1]: https://eprint.iacr.org/2017/1155.pdf section 4.1.3

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -99,7 +99,7 @@
 //! let my_sig_share = frost.sign(&x_only_frost_key, &session, my_index, &my_secret_share, my_nonce);
 //! # let sig_share3 = frost.sign(&x_only_frost_key, &session, party_index3, &secret_share3, nonce3);
 //! // receive the partial signature(s) from the other participant(s) and verify
-//! assert!(frost.verify_signature_share(&x_only_frost_key, &session, party_index3, sig_share3));
+//! assert_eq!(frost.verify_signature_share(&x_only_frost_key, &session, party_index3, sig_share3), Ok(()));
 //! // combine signature shares into a single signature that is valid under the FROST key
 //! let combined_sig = frost.combine_signature_shares(&x_only_frost_key, &session, vec![my_sig_share, sig_share3]);
 //! assert!(frost.schnorr.verify(
@@ -364,10 +364,8 @@ pub struct FrostKey<T: PointType> {
         ))
     )]
     public_key: Point<T>,
-    /// The image of each party's key share.
-    verification_shares: BTreeMap<PartyIndex, Point<Normal, Public, Zero>>,
-    /// Number of partial signatures required to create a combined signature under this key.
-    threshold: usize,
+    /// A threshold number of share images by their party index allowing us to reconstruct the full polynomial
+    core_shares: BTreeMap<PartyIndex, Point<Normal, Public, Zero>>,
     /// The tweak applied to this frost key, tracks the aggregate tweak.
     tweak: Scalar<Public, Zero>,
     /// Whether the secret keys need to be negated during signing (only used for EvenY keys).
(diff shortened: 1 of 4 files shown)


nickfarrow:
Still need some more time to go through this, thinking that because we don't *need* this feature just yet, and is turbo experimental, that we can leave for a later release?

schnorr_fun/src/frost.rs
```diff
@@ -892,38 +920,36 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
...
-            .expect("verifying party index that is not part of frost signing coalition")
+            .ok_or(VerifyShareError::PartyIsNotPartyOfSession)?
             .0;
         let R1 = R1.conditional_negate(session.nonces_need_negation);
         let R2 = R2.conditional_negate(session.nonces_need_negation);
-        g!(R1 + b * R2 + (c * lambda) * X - s * G).is_zero()
+        if g!(R1 + b * R2 + (c * lambda) * X - s * G).is_zero() {
+            Ok(())
+        } else {
+            Err(VerifyShareError::InvalidShare)
```
LLFourn, line 951: This should be `InvalidSignatureShare`.

LLFourn:
> Still need some more time to go through this, thinking that because we don't _need_ this feature just yet, and is turbo experimental, that we can leave for a later release?

The PR is mostly about improving the internals:

1. FROST key is serialized more compactly
2. Enrollment features can be added without changing serialization format.
3. There is one less panic case since you can always generate the verification share for any index.

So I think it's a *no-brainer* improvement. It's a bit less efficient computationally but I'd certainly take the above benefits over that.

I can remove the new API if it's too offensive but I actually think it's a good addition. For example it's the way you would recover the full key if you have only the seed words backing up 3 of the 5 shares. Perhaps I could rename it to `recover_key_from_share_images` to indicate it's only secure if you have first generated a key securely using the keygen protocol.

nickfarrow:
I think I am still having some difficulty conceptualizing the idea of core shares and how they will differ to non-core shares going forward (non-core shares also being points on the joint polynomial). 

The choice of evaluating the joint polynomial at the first `threshold` number of party indexes to be `core-shares` seems arbitrary? Having more than `t` core shares would interpolate the same joint polynomial. 

Core & non-core seem interchangeable for now, but is there a distinction going forward that I am missing? I think I understand this better in the case of enrollment where a party has not contributed any public poly during keygen and may have different "powers"/security.

But I have reviewed over the code and I agree it is a great improvement! Apart from my confusion re core vs non-core shares. Plus I mightve found a cool privacy application for enrollment 😎

schnorr_fun/src/frost.rs
```diff
@@ -438,12 +439,44 @@ impl FrostKey<Normal> {
...
+            core_shares: self.core_shares,
             tweak,
             needs_negation: self.needs_negation,
         })
     }
+
+    /// Creates a new FROST key from a set of core signer share images.
+    ///
+    /// The idea is that you can define the FROST key by a threshold number of key share images non-interactively.
+    /// Each scalar pre-image of teh `core_shares` will be the secret shares for this key.
```
nickfarrow, line 451: teh -> the

schnorr_fun/src/frost.rs
nickfarrow: Agreed! Maybe we will rename SignatureShare -> PartialSignature https://github.com/LLFourn/secp256kfun/issues/155

LLFourn:
> I think I am still having some difficulty conceptualizing the idea of core shares and how they will differ to non-core shares going forward (non-core shares also being points on the joint polynomial).
> 
> The choice of evaluating the joint polynomial at the first `threshold` number of party indexes to be `core-shares` seems arbitrary? Having more than `t` core shares would interpolate the same joint polynomial.

Strictly you can't guarantee that the `t+1` share is on the same polynomial as the original `t`. By only storing `t` you unconditionally guarantee that all the shares you have define a single polynomial. Maybe they could be called "basis shares" rather than 

Having said that I think this PR doesn't yet take this logic far enough. The public key itself can be derived from the core shares rather than storing it itself. I think I will try and think about this some more.

LLFourn:
I think the most efficient and flexible way of encoding the frost key is just the polynomial coefficients of the join polynomial. I don't think it should be too hard to reconstruct this from shares just by collecting the terms of the interpolation.  We shouldn't encode "tweaks" etc that should all be done by the application.

I might use this PR as a base to do the above.

LLFourn:
@nickfarrow wdyt about this feature now? The PR is no longer necessary but should be much easier to re-implement now with the tools you introduced. Is this needed as part of a backup workflow for the coordinator? Otherwise how does the coordinator restore the key when fist "syncing" with devices?

nickfarrow:
I think `t` should be the required threshold to load the key onto a new coordinator.

The only motivation I could think of was for verifying keygen, something like described in the FROST-DKG Equality Check.

Compared to their equality check protocol, involving FROST signing goes beyond a DKG. Signing with an overdetermined number of points sounds sketchy in this context. It could make implementation simpler.

PR #149, opened 2023-02-16
frost: Isolate creating shares and creating PoP

nickfarrow:
After discussion today, we do not need to know the other partys' polynomials in order to create their secret shares, PROVIDED we have some pre-existing way to assign the other parties an index
```
    pub fn create_shares(
        &self,
        party_indexes: impl Iterator<Item = usize>,
        scalar_poly: &Vec<Scalar>,
    ) -> Vec<Scalar<Secret, Zero>> {
```
```
    pub fn create_proof_of_posessions(
        &self,
        keygen_id: &[u8],
        scalar_poly: &Vec<Scalar>,
```

This PR is a step towards passing non-standard participant indexes into frost key generation as opposed to `1, 2, 3, ...`.
 
I also added   `create_shares_and_pop()` which behaves similar to the existing usage before this PR. 
I did this primarily because the iterator of indexes is not so friendly to pass in `1..=n_parties`, and we will soon be adding a sorting algorithm.

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -29,9 +29,9 @@
 //! let public_polys = vec![my_public_poly, public_poly2, public_poly3];
 //! let keygen = frost.new_keygen(public_polys).expect("something wrong with what was provided by other parties");
 //! // Generate secret shares for others and proof-of-possession to protect against rogue key attacks.
-//! let (my_shares, my_pop) = frost.create_shares(&keygen, my_secret_poly);
-//! # let (shares2, pop2) = frost.create_shares(&keygen, secret_poly2);
-//! # let (shares3, pop3) = frost.create_shares(&keygen, secret_poly3);
+//! let (my_shares, my_pop) = frost.create_shares_and_pop(&keygen, &my_secret_poly);
+//! # let (shares2, pop2) = frost.create_shares_and_pop(&keygen, &secret_poly2);
+//! # let (shares3, pop3) = frost.create_shares_and_pop(&keygen, &secret_poly3);
 //! // for i = 0..3, Send the secret share at index i and all proofs-of-possession to the participant with index i,
 //! // and receive our shares and pops from each participant as well.
 //! let received_shares = vec![my_shares[0].clone(), shares2[0].clone(), shares3[0].clone()];
@@ -246,6 +246,11 @@ impl KeyGen {
     pub fn n_parties(&self) -> usize {
         self.point_polys.len()
     }
+
+    /// Return the keygen id for this session
+    pub fn keygen_id(&self) -> [u8; 32] {
+        self.keygen_id
+    }
 }
 
 /// First round keygen errors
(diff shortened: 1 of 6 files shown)


schnorr_fun/src/frost.rs
```diff
@@ -25,13 +26,21 @@
 //! # let secret_poly3 = frost::generate_scalar_poly(threshold, &mut rng);
 //! # let public_poly2 = frost::to_point_poly(&secret_poly2);
 //! # let public_poly3 = frost::to_point_poly(&secret_poly3);
+//! let device_index = s!(1).public();
+//! # let device_index2 = s!(2).public();
+//! # let device_index3 = s!(3).public();
```
nickfarrow, line 31: @LLFourn I'd be interested in your thoughts on this API.

It's extra work for the user and this macro could be foreign to people.

However we can now enforce `type PartyIndex = Scalar<Public, NonZero>;` as the indexing parameters for a lot of functions. Now since our indexes start from 1 and we get rid of a bunch of `i+1`s in the polynomial/lagrange evaluations

I like it from an instructional point of view, evaluating the polynomials at some scalar index.

schnorr_fun/src/frost.rs
```diff
@@ -804,23 +884,23 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
...
                 .filter(|(j, _)| **j != index)
-                .map(|(j, _)| *j as u32 + 1)
+                .map(|(j, _)| *j)
                 .collect::<Vec<_>>(),
         );
         lambda.conditional_negate(frost_key.needs_negation);
         let c = &session.challenge;
         let b = &session.binding_coeff;
-        let X = frost_key.verification_shares().nth(index).unwrap();
+        let X = frost_key.verification_shares().get(&index).unwrap();
```
LLFourn, line 903: :bulb:  Moving to the "core signers" approach would get rid of this unwrap.

schnorr_fun/src/frost.rs
```diff
@@ -804,23 +884,23 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
         &self,
         frost_key: &FrostKey<EvenY>,
         session: &SignSession,
-        index: usize,
+        index: PartyIndex,
```
LLFourn, line 887: The "panics" doc is no longer correct on this function now I think.

schnorr_fun/src/frost.rs
```diff
@@ -437,27 +470,41 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
+            self.create_proof_of_posession(&keygen.keygen_id, scalar_poly),
+        )
+    }
+
+    /// Create our proof of posession to prove ownership of our scalar polynomial secret
+    ///
+    /// # Return value
+    ///
+    /// Returns a Signature of the keygen id under the first coeficient of our secret polynomial
+    pub fn create_proof_of_posession(&self, keygen_id: &[u8], scalar_poly: &[Scalar]) -> Signature {
```
LLFourn, line 501: Make this take a `schnorr_fun::Message`.

schnorr_fun/src/frost.rs
```diff
@@ -437,27 +470,41 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
...
     ///
-    /// ## Return value
+    /// # Return value
     ///
-    /// Returns a vector of secret shares where the index represents the signer index and a
-    /// proof-of-possession as a `Signature`.
-    pub fn create_shares(
+    /// Returns a vector of secret shares where the index represents the signer index,
+    /// and a proof of possession.
+    pub fn create_shares_and_pop(
```
LLFourn, line 478: This function doesn't make sense anymore because you need to pass the indexes you want to generate a share for. Return a `BTreeMap` of `PartyIndex` to share?

schnorr_fun/src/frost.rs
```diff
@@ -504,6 +551,20 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
         rng
     }
 
+    /// Create a keygen id to seed a keygen
+    /// TODO
+    pub fn create_keygen_id(&self, point_polys: &Vec<(Scalar<Public>, Vec<Point>)>) -> [u8; 32] {
```
LLFourn, line 556: why doesn't this just take a `KeyGen`? I see now that we are storing a `keygen_id` redundantly in `KeyGen`. It should just be `pub fn keygen_id(&self, keygen: &KeyGen) -> [u8;32]` and we shoudn't store it in `KeyGen`.

schnorr_fun/src/frost.rs
```diff
@@ -651,7 +732,7 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
     pub fn finish_keygen(
         &self,
         keygen: KeyGen,
-        my_index: usize,
+        my_index: PartyIndex,
         secret_shares: Vec<Scalar<Secret, Zero>>,
         proofs_of_possession: Vec<Signature>,
```
LLFourn, line 737: Need to pass in the message that the `proofs_of_possession` should be verified under.

schnorr_fun/src/frost.rs
```diff
@@ -611,22 +685,29 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
...
-            .map(|i| point_poly_eval(&joint_poly, (i as u32).into()).normalize())
+        // let mut keygen_hash = self.keygen_id_hash.clone();
+        // keygen_hash.update((len_first_poly as u32).to_be_bytes());
+        // keygen_hash.update((point_polys.len() as u32).to_be_bytes());
+        // for (_, poly) in &point_polys {
+        //     for point in poly {
+        //         keygen_hash.update(point.to_bytes());
+        //     }
+        // }
+        // let keygen_id: [u8; 32] = keygen_hash.finalize().into();
```
LLFourn, line 696: Remove commented out code.

LLFourn (changes requested):
I like this. We're clearly making things a bit harder to use by being less opinionated but as we are finding out it seems we need the flexibility.

nickfarrow:
TODO: Use BTreeMap in `new_keygen` and elsewere, preventing duplicate party indexes

LLFourn (approved):
I just did my own cleanup of this PR. Ready to merge with ACK from @nickfarrow.

PR #156, opened 2023-04-13
Do not rely on inference behavior for `Scalar::from_*` and `Point::from_*`

compiler-errors:
rust-lang/rust#110100 changes the inference behavior of the rust compiler subtly, which breaks the compilation of your crate. 

I attempted to remediate this by providing additional type hints to the compiler, but I'm not actually sure about these choices of `Secret` (or if they should be `Public`) since I know nothing about cryptography.


LLFourn:
Thanks for informing me of this!

LLFourn:
@compiler-errors would you mind checking that the changes I've made to this PR still compile with https://github.com/rust-lang/rust/pull/110100? Cheers.

compiler-errors:
Whoops, @LLFourn yes I tested and these new changes work fine.

PR #161, opened 2023-05-03
Fix dangerous Default impl for Point

### LINKED ISSUE #150: Point default is Zero while marked as NonZero

LLFourn:
You could get a zero point in a NonZero Point

fixes #150


nickfarrow (approved):

secp256kfun/src/point.rs
```diff
@@ -159,12 +172,35 @@ impl<T: PointType, S> Point<T, S, NonZero> {
...
+    where
+        T: PointType,
+    {
         let normalized = self.normalize();
         let needs_negation = !normalized.is_y_even();
         let negated = normalized.conditional_negate(needs_negation);
         (Point::from_inner(negated.0, EvenY), needs_negation)
     }
+
+    /// Returns the generator point [`G`] defined in [_Standards for Efficient Cryptography_].
```
nickfarrow, line 185: Docs don't like this [`G`]

PR #169, opened 2023-07-14
[frost] new_keygen safety

LLFourn:
Make it safer by encouraging passing in your secret polynomial directly.

### DIFF
--- a/schnorr_fun/src/frost.rs
+++ b/schnorr_fun/src/frost.rs
@@ -40,7 +40,7 @@
 //! ]);
 //! // (optionally) construct my_polys so we don't trust what's in public_poly_received for our index (in case it has been replaced with something malicious)
 //! let my_polys = BTreeMap::from_iter([(my_index, &my_secret_poly)]);
-//! let keygen = frost.new_keygen(public_polys_receievd, my_polys).expect("something wrong with what was provided by other parties");
+//! let keygen = frost.new_keygen(public_polys_received, &my_polys).expect("something wrong with what was provided by other parties");
 //! // Generate secret shares for others and proof-of-possession to protect against rogue key attacks.
 //! // We need pass a message to sign for the proof-of-possession. We choose the keygen
 //! // id here but anything works (you can even use the empty message).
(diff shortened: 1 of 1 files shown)


schnorr_fun/src/frost.rs
```diff
@@ -661,14 +663,34 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
...
         &self,
-        point_polys: BTreeMap<PartyIndex, Vec<Point>>,
-    ) -> Result<KeyGen, NewKeyGenError> {
+        mut point_polys: BTreeMap<PartyIndex, Vec<Point>>,
+        local_secret_polys: &BTreeMap<PartyIndex, S>,
+    ) -> Result<KeyGen, NewKeyGenError>
+    where
+        S: AsRef<[Scalar]>,
+    {
+        for (party_id, scalar_poly) in local_secret_polys {
```
nickfarrow, line 687: Is this futureproofing for having a single party control multiple shares of the FROST key?

schnorr_fun/src/frost.rs
```diff
@@ -586,12 +592,8 @@ impl<H: Digest<OutputSize = U32> + Clone, NG: NonceGen> Frost<H, NG> {
                 )
             })
             .collect::<BTreeMap<_, _>>();
-        let point_polys = scalar_polys
-            .iter()
-            .map(|(party_index, sp)| (*party_index, to_point_poly(sp)))
-            .collect::<BTreeMap<_, _>>();
 
-        let keygen = self.new_keygen(point_polys).unwrap();
+        let keygen = self.new_keygen(Default::default(), &scalar_polys).unwrap();
```
nickfarrow, line 596: 👍

nickfarrow (approved):
Looks good!

This risk was introduced by us being lazy not passing our public poly into finish keygen code. Just receiving a `FinishKeygen` message from the coordinator which contains all the public polys including our own.

Others will likely do the same, so definitely worth highlighting.

It does make `new_keygen` slightly clunkier, skipping the check with `BTreeMap::new() or Default::default()`, but i think this is probably nicer than an `Option<BTreeMap<_, _>>`

nickfarrow:
Looks good!

This risk was introduced by us being lazy not passing our public poly into finish keygen code. Just receiving a `FinishKeygen` message from the coordinator which contains all the public polys including our own, 

Others will likely do the same, so definitely worth highlighting.

It does make `new_keygen` slightly clunkier, skipping the check with `BTreeMap::new() or Default::default()`, but i think this is probably nicer than an `Option<BTreeMap<_, _>>`

schnorr_fun/src/frost.rs
LLFourn: No it allows you to contribute multiple secret polynomials. Should probably put on the docs that it doesn't make sense to design a key genration protocol where you do this.

LLFourn:
Thanks @nickfarrow.

> It does make `new_keygen` slightly clunkier, skipping the check with `BTreeMap::new() or Default::default()`, but i think this is probably nicer than an `Option<BTreeMap<_, _>>`

Clunkyness is a feature as it inspires the reader to look at the documentation!

PR #170, opened 2023-09-19
Use procedural macro for `g!` and `s!` macros

### LINKED ISSUE #125: Make API of op::lincomb not require collecting first
see #116

### LINKED ISSUE #146: Make everyghing `Copy`
See discussion: https://github.com/LLFourn/secp256kfun/issues/6#issuecomment-1363752651

This might also help fix: https://github.com/LLFourn/secp256kfun/issues/125?

LLFourn:
This gives us much more expressivity in the macros:

1. No more `{..}` when you need to call methods or access tuple or array elements.
2. New `.*` dot product operator which replaces manually called `op::lincomb`.
3. New scalar division operator `/` (mostly for fun -- it's used in one place in ecdsa_fun).

RIP macro_rules stack machine.

In order to design the thing correctly I also tried to make everything copy (#146) to maintain backwards compat with omitting &.
In the end I didn't really use that but instead just allowed you to pass a reference to most of the functions in op which also fixes #125.

This is probably a breaking change due to some slight changes in the way some parts of the expressions are interpreted.

I've done benchmarks to confirm that there's no performance regression in terms of what is produced.

Fixes #146 #125 #6

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,7 @@ on:
 # Make sure CI fails on all warnings, including Clippy lints
 env:
     RUSTFLAGS: "-Dwarnings"
+    RUSTDOCFLAGS: "-Dwarnings"
 
 jobs:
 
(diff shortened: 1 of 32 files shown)


nickfarrow:
Changes look great to me, looked at FROST changes specifically and skimmed most of the other code (though i'm lacking skills to to review the macros themselves yet)

LLFourn:
Hey I added some notes on how I would structure the "problem we are solving" part of the pitch which compares existing solutions to ours.

PR #171, opened 2023-12-18
[frost] backup scheme

nickfarrow:
Explanation and justification of scheme in module documentation.

This draft scheme currently includes the ability for the share index to either be a small integer or a scalar. 

- [x] draft scheme
- [ ] figure out what is going on with checksum length and warn about error-correction if necessary.
- [x] encode to bech32m and decode
- [x] check ive implemented optional `share_backup` feature and bech32 dependency correctly
- [x] restructure FROST internally to use polynomials
- [x] more comprehensive tests

frostsnap-cli backup and restore (separate PR)


schnorr_fun/src/frost_backup.rs
```diff
@@ -0,0 +1,198 @@
...
+///
+/// Requires that the threshold be no greater than 32.
+/// If using an integer (small) index, the backup will be 67 bech32 characters.
+/// If using a scalar index, the backup will be 118 bech32 characters.
+pub fn encode_backup<H: Default + Digest<OutputSize = U32>>(
+    threshold: usize,
+    polynomial: Vec<Point>,
+    secret_share: Scalar<Secret>,
+    share_index: BackupShareIndex,
+) -> Result<String, FrostBackupError> {
```
LLFourn, line 107: Encoding and decoding errors should be different. I don't know if there is any justifiable reason to have an error for encoding. Maybe we want to make sure the threshold fits but everything else seems like a programmer error (can just be a panic). Rather than return an error you could have a `ShortThreshold` type you force the caller to construct it to make them better aware of the constraint. Or `NonZeroU5` since you can also use it for the party index.

schnorr_fun/src/frost_backup.rs
```diff
@@ -0,0 +1,198 @@
...
+/// The first 4 bech32 chars from a hash of the polynomial coefficients.
+/// Collision expected once in (32)^4 = 2^20.
+pub fn polynomial_identifier<H: Default + Digest<OutputSize = U32>>(
+    polynomial: Vec<Point>,
+) -> [u5; 4] {
+    let mut hash = H::default();
+    for coefficient in polynomial.into_iter() {
+        hash.update(coefficient.to_bytes());
+    }
+    hash.finalize().to_vec().to_base32()[0..4]
```
LLFourn, line 92: The  `to_vec` seems unnecessary. Can't you call `to_base32` on the result of finalize?

I don't think this needs to be a separate function either?
I think you can do `hash.add(&polynomial[..])` if you import the `HashAdd` trait rather than do a loop.

schnorr_fun/src/frost_backup.rs
```diff
@@ -0,0 +1,198 @@
...
+};
+
+/// Secret shares can have a small index up to 32, or index at a specific scalar.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum BackupShareIndex {
+    /// a secret share at a small index up to 32
+    SmallIndex(usize),
+    /// a secret share at a specific scalar index
+    Scalar(Scalar),
+}
```
LLFourn, line 59: Maybe just always use a scalar and check if you can use the short encoding?

schnorr_fun/src/frost_backup.rs
```diff
@@ -0,0 +1,198 @@
...
+//! We could also leave higher thresholds for a later backup version.
+//!
+//! ### Polynomial Identifier - (4)
+//! Hash of the polynomial coefficients and take the first 4 bech32 characters.
+//! This identifier will allow determination of secret share compatibility.
+//!
+//! It is possible to have two different polynomials that interpolate the same joint-secret,
+//! but shares on these different polynomials will not be compatible with one another.
+//!
+//! 32^4 = 2^20 ~> a collision rate of about one in a million.
```
LLFourn, line 34: To be precise the probability of two different polynomials colliding is ~1 in a million.

schnorr_fun/tests/frost_backup.rs
```diff
@@ -0,0 +1,75 @@
...
+            "34f7ce653cfa8454b3463726a599ef2925736442d2d06455974d6feae9450d90",
+        )
+        .unwrap(),
+    );
+
+    let frost_backup = encode_backup::<sha2::Sha256>(
+        threshold,
+        polynomial.clone(),
+        secret_share.clone(),
+        share_index.clone(),
```
LLFourn, line 60: These clones shouldn't be necessary. You can change the API to take references.

schnorr_fun/src/frost_backup.rs
```diff
@@ -0,0 +1,198 @@
...
+    let threshold_u5 =
+        u5::try_from_u8((threshold - 1).to_le_bytes()[0]).map_err(FrostBackupError::Bech32Error)?;
+    data[0] = threshold_u5;
+
+    let polynomial_identifier_u5 = polynomial_identifier::<H>(polynomial);
+    for (i, byte) in polynomial_identifier_u5.into_iter().enumerate() {
+        data[1 + i] = byte;
+    }
+
+    let secret_share_u5 = secret_share.to_bytes().to_vec().to_base32();
```
LLFourn, line 124: should we check the share is correct with respect to the polynomial?
Perhaps we should have an encode variant that does that.

LLFourn (commented):
1. Given a vector of shares we should have a function that tries to reconstruct the main secret.
2. Given a vector of share images should try and reconstruct the public polynomial and therefore give us the ability to check the id in the encoding.

schnorr_fun/src/frost_backup.rs
```diff
@@ -0,0 +1,198 @@
+//! Backup scheme for FROST secret shares
+//!
+//! # Description
+//!
+//! Based on https://bips.xyz/93
+//!
+//! ## Backup format (bech32 chars)
+//!
+//! human readable: "fr"            // (2)
```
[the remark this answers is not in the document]
nickfarrow: I agree with the premise but think using `fr` to indicate a frost backup is cooler to users

SSS plebs can use frost labelled backups and reconstitute their seed if they must 😄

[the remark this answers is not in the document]
nickfarrow: Is our specification going to be specifically for frost backups, or a wider shamir secret sharing backup spec? 

If we don't capture a wider SSS specification here and one was made in the future, it will likely be able to say `fr` data is compatible with `sss`.

And if it wasn't compatible, then this implies a wider SSS backup spec needs data that we are missing currently. I don't think there is anything extra needed, but we ourselves are not using SSS for anything, and maybe someone wants to do something clever in the future.

`sss` is also longer, and `ss` is no good

schnorr_fun/src/frost_backup.rs
nickfarrow: True it is now so simple it doesn't really need to be a separate function, but it might make things just a little bit easier for users of our FROST implementation, easily check their polynomial matches a backup using `polynomial_identifier()`.

& useful to separate for a spec

nickfarrow:
I've also taken this as a good time to create a separate module for polynomials `secp256kfun::poly`.
By including all the lagrange interpolation utilities in here, the secret share backup scheme does not depend on `frost.rs`.

I made the `FrostKey` store the public polynomial internally, and use it to generate verification shares and threshold.

Two things of interest:

We lose sight of which signer indexes exist internally, and also `pub fn n_signers(&self) -> usize`. We don't use it anywhere inside our logic, so does this matter? It might mean that users of this library will need to store signers & number of signers separately alongside the `FrostKey`. This kind of makes sense as we abstract away from a FrostKey having a fixed number of signers.

Secondly, I am wondering if we want to go further and remove storing the FROST `public_key` as we are currently? Since it changes with tweaks depending on the FrostKey point-type, it strays from `point_polynomial[0]`. Or maybe we just rename it to `tweaked_public_key`.

Perhaps we could do something like accumulate these tweaks internally until `.public_key()` is called. I recall we didn't like the API of another implementation which took vecs of ordinary & xonly tweaks as arguments. Do you have any thoughts on this or our current `FrostKey<T>` API?

LLFourn:
> I've also taken this as a good time to create a separate module for polynomials `secp256kfun::poly`. By including all the lagrange interpolation utilities in here, the secret share backup scheme does not depend on `frost.rs`.

Cool.

> 
> I made the `FrostKey` store the public polynomial internally, and use it to generate verification shares and threshold.
> 
> Two things of interest:
> 
> We lose sight of which signer indexes exist internally, and also `pub fn n_signers(&self) -> usize`. We don't use it anywhere inside our logic, so does this matter? It might mean that users of this library will need to store signers & number of signers separately alongside the `FrostKey`. This kind of makes sense as we abstract away from a FrostKey having a fixed number of signers.

Yes this is the way to go.

> 
> Secondly, I am wondering if we want to go further and remove storing the FROST `public_key` as we are currently? Since it changes with tweaks depending on the FrostKey point-type, it strays from `point_polynomial[0]`. Or maybe we just rename it to `tweaked_public_key`.

We shouldn't store on disk but we should have a separate public key field in memory. `tweaked_public_key` SGTM.

> 
> Perhaps we could do something like accumulate these tweaks internally until `.public_key()` is called. I recall we didn't like the API of another implementation which took vecs of ordinary & xonly tweaks as arguments. Do you have any thoughts on this or our current `FrostKey<T>` API?

I think the current API is good.

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,260 @@
...
+/// A vector with a tail of zero coefficients means the interpolation was overdetermined.
+pub fn interpolate_point_polynomial(
+    indexes: Vec<Scalar<impl Secrecy, impl ZeroChoice>>,
+    points: Vec<Point>,
+) -> Vec<Point<impl PointType, Public, Zero>> {
+    // Get each lagrange basis polynomial from the product of coefficients:
+    //      l_j(x) = Product[ (x-x_m)/(x_j-x_m), j!=m]
+    // Or
+    //      l_j(x) = Product[ a_m*x + b_m, j!=m], where a_m = 1/(x_j-x_m) and b_m = -x_m*a_m.
+    let basis_polynomials: Vec<_> = indexes
```
LLFourn, line 94: I think this would be easier to understand as a single `for` loop that starts with `sum` an identity poly (all Point::zero) and figures out the basis poly in each iteration and adds the basis poly to `sum`.

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,260 @@
...
+        .clone()
+        .into_iter()
+        .enumerate()
+        .map(|(j, x_j)| {
```
LLFourn, line 98: I think this bit of the map could go into its own function like:

```rust
fn unit_basis_poly(indices: &[Scalar<Public, impl ZeroChoice>] ) -> Vec<Vec<Scalar<Public, Zero>>>
```
I call this the unit basis because the basis is the one after you multiply by `y` I think. There's probably a better term for this. Not sure if it's a good idea.

secp256kfun/src/poly.rs
nickfarrow: Do you mean sum the basis polynomials as we go? I think this would be more confusing, I like how it is broken out into steps: find unit basis polys, multiply to intersect points, then sum. Or do you mean something else?

schnorr_fun/src/frost_backup.rs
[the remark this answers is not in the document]
nickfarrow: changed to `frost`

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,260 @@
...
+            basis_poly
+                .iter()
+                .map(|coeff| g!(coeff * y_value))
+                .collect::<Vec<_>>()
+        })
+        .collect();
+
+    let mut point_polynomial = interpolating_basis[0].clone();
+    for poly in &interpolating_basis[1..] {
+        for (i, point) in poly.iter().enumerate() {
```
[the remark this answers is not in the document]
nickfarrow: I've added `poly::add` for combining basis point polynomials, and I've added `poly::point_mul` for multiplying a the scalar unit basis polynomials to interpolate a point

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+) -> Vec<Point<NonNormal, Public, Zero>> {
+    poly.iter()
+        .map(|coeff| g!(coeff * point).mark_zero())
+        .collect::<Vec<_>>()
+}
+
+/// Add the coefficients of two point polynomials.
+///
+/// Handles mismatched polynomial lengths.
+pub fn add<T: PointType + Default, S: Secrecy, Z: ZeroChoice>(
```
LLFourn, line 161: this should be in the `point` module also.

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+        .collect()
+}
+
+/// Interpolate a set of shamir secret shares to find the joint secret.
+///
+/// Each shamir secret share is associated with a participant index (index, share).
+///
+/// Panics if the indicies are not unique.
+pub fn reconstruct_shared_secret(
+    secrets_at_indices: Vec<(Scalar, Scalar<impl Secrecy, impl ZeroChoice>)>,
```
LLFourn, line 189: surely the scalars are secret here!

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+    poly: &[Scalar],
+    x: Scalar<impl Secrecy, impl ZeroChoice>,
+) -> Scalar<Secret, Zero> {
+    s!(powers(x) .* poly)
+}
+
+/// Evaluate a point polynomial defined by coefficients, at some index.
+///
+/// The polynomial coefficients begin with the smallest degree term first (the constant).
+pub fn point_poly_eval(
```
LLFourn, line 53: I still think it's better separate operations on `&[Scalar]` and `&[Point]` in different modules. It can be in this file just put `mod point { ... }` around the stuff and remove `point` from the function names (and do the same for scalars).

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+        .collect();
+
+    interpolating_basis.into_iter().fold(
+        Vec::with_capacity(basis_polynomials[0].len()),
+        |acc, poly| add(&acc, &poly),
+    )
+}
+
+/// Multiply each coefficient in a scalar polynomial by a point.
+pub fn point_mul(
```
LLFourn, line 149: I'd remove and inline this one. Perhaps I made a complaint that caused you to add it. I was probably confused that this was a single scalar against a point polynomial rather than the other way around. Multiplying a scalar vector by a point is fairly incoherent conceptually (it's usually the other way around!).

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+///
+/// A vector with a tail of zero coefficients means the interpolation was overdetermined.
+pub fn interpolate_point_polynomial(
+    points_at_indicies: Vec<(Scalar<Public, impl ZeroChoice>, Point)>,
+) -> Vec<Point<impl PointType, Public, Zero>> {
+    let (indicies, points): (Vec<_>, Vec<_>) = points_at_indicies.into_iter().unzip();
+
+    let basis_polynomials: Vec<_> = unit_basis_polys(indicies.as_slice());
+
+    let interpolating_basis: Vec<_> = basis_polynomials
```
LLFourn, line 136:
```suggestion
    let point_scaled_basis_polynomials: Vec<_> = basis_polynomials
```

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+        })
+        .collect();
+    secrets
+        .into_iter()
+        .zip(coefficients)
+        .fold(s!(0).mark_zero(), |acc, (secret, coefficient)| {
+            s!(acc + secret * coefficient)
+        })
+        .non_zero()
+        .expect("joint secret should not be zero")
```
LLFourn, line 208: I think we should definitely allow it be zero here.

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+    let coefficients: Vec<_> = indicies
+        .iter()
+        .map(|my_index| {
+            lagrange_lambda(
+                *my_index,
+                indicies.clone().into_iter().filter(|j| j != my_index),
+            )
+        })
+        .collect();
+    secrets
```
LLFourn, line 201: This doesn't need to do two iterations. You can just multiply in the first iteration.

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,323 @@
...
+        .into_iter()
+        .zip(coefficients)
+        .fold(s!(0).mark_zero(), |acc, (secret, coefficient)| {
```
LLFourn, line 204: I don't think you need to mark 0 as zero :)

secp256kfun/src/poly.rs
LLFourn: > I like how it is broken out into steps: find unit basis polys, multiply to intersect points, then sum

Summing the scaled basis polynomials as we go is also broken up into steps. You figure out the basis poly, scale by the y coordinate and then add it back to the sum. Each step seems simpler also because it doesn't involve an iteration. I think implementing an algorithm in the fewest number of iterations usually leads to the simplest to understand code and maps to the textbook reading of the algorithm (there is only one loop):

image

This is just a point of style and not a huge concern for me so just roll with what you feel. I understand that every rust programmer has to go through the use successive iterator combinators everywhere stage of their career!

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,260 @@
...
+        .into_iter()
+        .enumerate()
+        .map(|(j, x_j)| {
+            let mut coefficients: Vec<_> = vec![];
+            for (_, x_m) in indexes.iter().enumerate().filter(|(m, _)| *m != j) {
+                let a_m = s!(x_j - x_m)
+                    .non_zero()
+                    .expect("points must lie at unique indexes to interpolate")
+                    .invert();
+                let b_m = s!(-x_m * a_m).mark_zero();
```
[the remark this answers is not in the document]
nickfarrow: I didn't use it here because simplest form in the product was $(a_m * x + b_m)$ terms rather than $(x / a_m + b_m)$, also yeah to use `/` here you'd have to do something like 
```
let a_m_denom = s!(x_j - x_m)
    .non_zero()
    .expect("points must lie at unique indicies to interpolate");
let a_m = s!(1 / a_m_denom);
```

schnorr_fun/src/frost.rs
```diff
@@ -186,11 +186,20 @@ use secp256kfun::{
...
-type PartyIndex = Scalar<Public, NonZero>;
+/// The index of a party's secret share.
+///
+/// This index and its secret share define a point on the joint secret polynomial.
+/// It is used in interpolation and computation of the shared secret.
+///
+/// This index can be any non-zero [`Scalar`], but must be unique between parties.
+/// In most cases it will make sense to use simple indicies `s!(1), s!(2), ...` for smaller backups.
+/// Other applications may desire to use indicies corresponding to pre-existing keys or identifiers.
+pub type PartyIndex = Scalar<Public, NonZero>;
```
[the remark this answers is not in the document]
nickfarrow: Can i add `--all-features` to our doc test: `cargo doc --no-deps --workspace`?

secp256kfun/src/poly.rs
nickfarrow: Ah got you, changed. Better?

secp256kfun/src/poly.rs
nickfarrow: Also adding `interpolate_point_polynomial` to `poly::point::interpolate`

schnorr_fun/src/share_backup.rs
```diff
@@ -0,0 +1,238 @@
...
+        return Err(FrostBackupDecodeError::InvalidHumanReadablePrefix);
+    }
+
+    if !matches!(variant, bech32::Variant::Bech32m) {
+        return Err(FrostBackupDecodeError::WrongBech32Variant(variant));
+    }
+
+    let mut threshold_bytes =
+        Vec::<u8>::from_base32(&data[..2]).map_err(FrostBackupDecodeError::Bech32DecodeError)?;
+    threshold_bytes.resize((usize::BITS / 8) as usize, 0);
```
nickfarrow, line 199: I had to do this to work on 32bit architectures, reason enough to use something instead of usize for threshold?

nickfarrow:
One thing I haven't implemented here is having an `EncodableFrostKey` that doesnt serialize the tweaked public key

schnorr_fun/src/frost.rs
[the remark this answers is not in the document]
LLFourn: yes!

schnorr_fun/src/share_backup.rs
LLFourn: This should just be a u16

schnorr_fun/src/share_backup.rs
```diff
@@ -0,0 +1,238 @@
...
+    };
+
+    bech32::encode("frost", &data[..(data.len() - n_unused_bytes)], Bech32m)
+        .expect("hrp must be valid")
+}
+
+/// Decode a bech32m secret share backup
+pub fn decode_backup(
+    encoded: String,
+) -> Result<(usize, [u5; 4], Scalar, Scalar), FrostBackupDecodeError> {
```
LLFourn, line 185: Make a struct for the return value here so it's easier to tell what each field means. My plan would be:

1. Make a struct `ShareBackup` implement display on it which should product the actual string.
2. When you decode the string you get `ShareBackup`
3. You have  a `ShareBackup::new` which takes the same arguments as the encode function now.
4. You can also have a floating encode function for convenience too.

I know you originally had something more like this and I suggested single functions. I didn't realize at that time the decode return value would be so complicated. Sorry!

nickfarrow:
Added `ShareBackup` struct, I think this API is nice!

There's a duplicated `threshold < 1024` because of the floating encode but i think this is fine.

Reminder that I'd like to squash these commits after any approval.
I think I like this flow of keeping working commits separate for easy review, and then squashing appropriately so our master looks perfect.

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,317 @@
+//! Utilities for working with polynomials on the secp256k1 elliptic curve.
+//!
+//! A polynomial f(x) of degree k is defined by its coefficients
+//!
+//! `f(x) = a_0 + a_1 * x + a_2 * x^2 + ... + a_k * x^k`
```
LLFourn, line 5: defined by its coefficients `a_0, ... a_n`. The coefficients can be `Scalar` or `Point`.

I wonder if you can use LaTeX in rust doc yet.

LLFourn (commented):
Left every nit I can find. Looks really really god.

> One thing I haven't implemented here is having an EncodableFrostKey that doesnt serialize the tweaked public key

Please do this and we'll be done here.

secp256kfun/src/poly.rs
nickfarrow: funnily enough search took me to the zcash frost squad discussing using latex in their rust doc https://github.com/ZcashFoundation/frost/pull/436#discussion_r1265895682, apparently you can using mathjax in html header..

schnorr_fun/tests/share_backup.rs
```diff
@@ -0,0 +1,58 @@
...
+        decoded_share_backup.identifier
+    );
+    assert_eq!(secret_share, decoded_share_backup.secret_share);
+    assert_eq!(share_index, decoded_share_backup.share_index);
+}
+
+#[test]
+fn share_backup_long() {
+    let secret_poly = vec![s!(1), s!(2), s!(3)];
+    let polynomial: Vec<_> = secret_poly.iter().map(|c| g!(c * G).normalize()).collect(); // some polynomial coefficients
```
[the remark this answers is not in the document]
nickfarrow: :man_facepalming:

secp256kfun/src/poly.rs
```diff
@@ -0,0 +1,317 @@
...
+        s!(acc * x_m / denominator).public()
+    })
+}
+
+/// Interpolate a set of shamir secret shares to find the joint secret.
+///
+/// Each shamir secret share is associated with a participant index (index, share).
+///
+/// Panics if the indicies are not unique.
+pub fn reconstruct_shared_secret(
```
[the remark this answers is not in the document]
nickfarrow: I've moved into `mod scalar` but didn't make a similar function in `mod point`

nickfarrow:
Please check what i've done with `EncodedFrostKey`, is this what you had in mind? It only stores the point polynomial, from which an untweaked `FrostKey` can be restored. I guess we don't even want to store the Secret Share here because we'll be using better security for that.

I've removed the serde and bincode stuff from `FrostKey`

nickfarrow:
I was just exploring a test that randomly deletes `n` random characters from the backup. It doesn't seem the `bech32` is doing the error correction we would like it to, I just get `Bech32DecodeError(bech32::Error::InvalidChecksum)` even when deleting a single character. 

Sure the checksum is invalid, but i'd think for bech32 to be useful we want to see the corrected decoding suggestion(s)? `InvalidChecksum(PotentialDecoding)`

schnorr_fun/src/share_backup.rs
```diff
@@ -0,0 +1,319 @@
...
+pub fn encode_backup(
+    threshold: u16,
+    identifier: [u5; 4],
+    secret_share: Scalar,
+    share_index: Scalar<Public>,
+) -> String {
+    if threshold == 0 {
+        panic!("Polynomial threshold can not be zero");
+    }
+    if threshold > 1024 {
```
LLFourn, line 183: These panics are duplicated in `ShareBackup::new` can we design this so they are only done in one spot? I think this function can probably be deleted now that we have `ShareBackup`.

Also it's better to use `assert!` and `assert_ne!` here because they give slightly nicer errors with less effort.

LLFourn (approved):
LGTM. Thanks.

I fixed up one thing. The decode/deserialize impls for `EncodedFrostKey` should not allow zero as the first coefficient.

schnorr_fun/src/frost.rs
```diff
@@ -1066,50 +1025,45 @@ where
...
+#[cfg_attr(
+    feature = "serde",
+    derive(crate::fun::serde::Deserialize, crate::fun::serde::Serialize),
+    serde(crate = "crate::fun::serde")
+)]
+#[cfg_attr(
+    feature = "bincode",
+    derive(crate::fun::bincode::Encode, crate::fun::bincode::Decode),
+    bincode(crate = "crate::fun::bincode",)
+)]
```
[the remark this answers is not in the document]
LLFourn: (update: I did this)

PR #175, opened 2024-02-13
[frost] EncodedFrostKey BorrowDecode and correct markers

nickfarrow:
- Implement `BorrowDecode` for `EncodedFrostKey` for use in Frostsnap.
- Convenient helper impls for `EncodedFrostKey`.
    * I was finding I had to call things like `Into::<FrostKey<Normal>>::into(session_progress.encoded_key.clone())` everywhere, `.into_frost_key()` is nicer.
- Fully allow zero polynomial coeffs in share_backup (mistakenly missed a few functions last PR).
- Keygen local poly Scalars should be secret.


schnorr_fun/src/frost.rs
```diff
@@ -1092,6 +1104,15 @@ impl crate::fun::bincode::Decode for EncodedFrostKey {
     }
 }
 
+#[cfg(feature = "bincode")]
+impl<'de> crate::fun::bincode::BorrowDecode<'de> for EncodedFrostKey {
```
LLFourn, line 1108: Did you try this: https://docs.rs/bincode/2.0.0-rc.3/bincode/macro.impl_borrow_decode.html?

schnorr_fun/tests/share_backup.rs
```diff
@@ -6,7 +6,10 @@ use secp256kfun::{marker::Secret, poly, s, Scalar};
 #[test]
 fn short_backup() {
     let secret_poly = vec![s!(6), s!(1), s!(9)];
-    let polynomial: Vec<_> = poly::scalar::to_point_poly(&secret_poly); // some polynomial coefficients
+    let polynomial: Vec<_> = poly::scalar::to_point_poly(&secret_poly)
+        .into_iter()
+        .map(|c| c.mark_zero())
+        .collect(); // some polynomial coefficients
```
LLFourn, line 12: why do we need all this map stuff now?

(aside: I think `// some polynomial coefficients` is not doing much here)

schnorr_fun/src/frost.rs
```diff
@@ -694,7 +694,7 @@ impl<H: Digest<OutputSize = U32> + Clone, NG> Frost<H, NG> {
         local_secret_polys: &BTreeMap<PartyIndex, S>,
     ) -> Result<KeyGen, NewKeyGenError>
     where
-        S: AsRef<[Scalar]>,
+        S: AsRef<[Scalar<Secret>]>,
```
LLFourn, line 697: `Scalar` is `Secret` by default.

LLFourn:
ConceptACK

LLFourn:
I've pushed two commits:

1. 3435eae4d1a2bb5432d73d68096ab43c6c59f323 to bump the MSRV so tests pass
2. e6e217e4b05bcebd6ee93bb7fead1daf69b1d3da to set the secret share to `Scalar<Secret, Zero>` and fix some misc things

One thing I noticed is that the shares index can be one u5 or 52 but nothing inbetween. I think it should be fine to allow it to be any length. I guess @nickfarrow didn't do this because it was tricky with the API (there is this `convert_bits` thing but it seems to be broken). It looks like it might be possible with the `v0.10` branch.

PR #182, opened 2024-05-13
More concise share backups

nickfarrow:
Exploring smaller backups by removing polynomial identifier and threshold (should be written in plaintext alongside the bech32 backup).


LLFourn:
@nickfarrow ready for your review. Review each commit independently is probably best.

secp256kfun/src/poly.rs
```diff
@@ -45,25 +46,92 @@ pub mod scalar {
...
+    }
+
+    /// Adds two scalar polynomials together where `apoly` is mutated to form the result.
+    pub fn add_in_place<SA: Secrecy>(
+        apoly: &mut Vec<Scalar<SA, Zero>>,
+        bpoly: impl IntoIterator<Item = Scalar<impl Secrecy, impl ZeroChoice>>,
+    ) {
+        let b = bpoly.into_iter();
+        for (i, b) in b.enumerate() {
+            if i == apoly.len() {
```
nickfarrow, line 113: at first glance i thought this was wrong because of the `==` for unequal length polys, but the following push makes it repeat this branch on the next iteration :+1:

nickfarrow:
> frost shares should be Zero 

Could you explain this one to me, why allow a frost secret share to be zero?

While zero isnt inherently forbidden for secret shares, it should never occur from a secure keygen, only if someone is cancelling out secret poly(s) and that might be worth panicking (perhaps not in this way!). I want to understand when to make things 'illegal', only when it's actually illegal within that context (like `send_pubkey_to_bob` example in readme)?

schnorr_fun/src/frost/share.rs
```diff
@@ -0,0 +1,297 @@
+use secp256kfun::{marker::*, poly, Scalar};
```
nickfarrow, line 1: nit: any reason to put this `use` before the docs?

nickfarrow: Ah yes it's documentation on SecretShare

schnorr_fun/src/frost/share.rs
```diff
@@ -0,0 +1,297 @@
...
+            test_runner::{RngAlgorithm, TestRng},
+        },
+        G,
+    };
+    proptest! {
+        #[test]
+        fn recover_secret(parties in 1usize..10, threshold in 1usize..5) {
+            use rand::seq::SliceRandom;
+            let frost = frost::new_with_deterministic_nonces::<sha2::Sha256>();
+            let parties = parties.max(threshold);
```
nickfarrow, line 288: could use `(parties, threshold) in (1usize..=10).prop_flat_map(|n| (Just(n), 1usize..=n))` but it's also kind of confusing

LLFourn:
> > frost shares should be Zero
> 
> Could you explain this one to me, why allow a frost secret share to be zero?
> 
> While zero isnt inherently forbidden for secret shares, it should never occur from a secure keygen

This is true for every value that `Scalar` can take. The easy way to tell then is if your concern about `0` would equally apply to `42` or `SHA256("malicious key")`, then there's no problem with 0.

LLFourn (approved):

Issue #189, opened 2024-07-22
`FrostKey::mul_tweak`


irriden:
If we ask signers to return `s!(r1 + (r2 * b) + lambda * x * c * mul_tweak)` in `Frost::sign`, we can get a signature that is valid for the public key `s * mul_tweak * G`.

Any multiplicative tweaks would have to be applied before any additive tweaks.

Let me know if you would merge a PR that implements this, thank you!

LLFourn:
Hi irriden,

I have been talking about going in a direction with "tweaking" that would tweak the secret shares as well as the public `Frostkey`. This means `FrostKey` would no longer "store" tweaks but would mutate the polynomial by the tweaks. In the case of addition this would mean just adding the value to the first coefficient. In the case of multiplication it would be multiplying the entire polynomial by the public scalar. You of course have to do the homomorphic operation to the shares as well. This would allow you to do what you want. What do you think?

What do you actually want this for out of interest.

LLFourn:
> I am researching multisig lightning. In lightning the `revocationpubkey` for a specific commitment is derived using a multiplicative, followed by an additive tweak.

Oh ok weird lol -- why? I can put multiplication into the API that I am working on right now. I think it's sound from a security point of view. To be clear you wouldn't need to keep the polynomial in the new API -- you can tweak your secret share if you're a signer. `verify_signature_share` will be the thing that still needs it.

I'm guessing this is not "xonly" tweaking? The resulting point does not have its y coordinate dropped?

I am about to start working on making it possible to do `bip-frost-dkg` compatible ops for what it's worth.

LLFourn:
> See here for details: https://github.com/lightning/bolts/blob/master/03-transactions.md#revocationpubkey-derivation

Thanks. I wonder why that needed to be a multiplication.
 
> > `verify_signature_share` will be the thing that still needs it.
> 
> Would you consider having `verify_signature_share` depend not on the polynomial, but on the public share of the corresponding signer, as in `bip-frost-dkg` ? Thank you.

Hmmm I certainly could add an API for this -- weirdly enough we do have a `verification_share` method to get it but nothing to use it :facepalm:. The problem with verification shares is that their value changes when a key is tweaked. How do you handle this right now? Do you multiply the verification shares when you are doing a multiplicative tweak? Or is this being stored in some separate state and being patched up in the verification check. I feel like if we make a verification share based API it will be very error prone because there is yet another thing to keep track off the tweaks on.

[...]

PR #190, opened 2024-07-25
[frost] Make "tweaks" mutate

LLFourn:
Prior to this PR tweaks were being stored in a separate field inside `FrostKey` and were applied at signature combination time. In frostsnap we actually want to be able to do "one-way" tweaks where you forget the original polynomial (at least the first coefficient). This is not an API safety concern but we strictly don't want the original to exist anymore.

The next problem we address is that we don't want you to need the full polynomial for signing. Instead we make the user provide a "paired" secret share which is the secret scalar and index paired with the shared key (first coefficient of polynomial). This is a convenient combination since it allows you to hash the shared key to figure out a tweak and then apply it to both the shared key and the shared secret.

See commits individually. eb6dc14994d0dc99e1b1118d58d7128aff386ca5 is the frost only one.

I haven't updated docs yet because I intend to integrate some bip-frost stuff which will change the API again.

### DIFF
--- a/schnorr_fun/src/frost/share.rs
+++ b/schnorr_fun/src/frost/share.rs
@@ -1,4 +1,4 @@
-use secp256kfun::{marker::*, poly, Scalar};
+use secp256kfun::{poly, prelude::*};
 /// A *[Shamir secret share]*.
 ///
 /// Each share is an `(x,y)` pair where `y = p(x)` for some polynomial `p`. With a sufficient
@@ -55,7 +55,7 @@ pub struct SecretShare {
     /// value (other than 0).
     pub index: Scalar<Public>,
     /// The secret scalar which is the output of the polynomial evaluated at `index`
-    pub secret: Scalar<Secret, Zero>,
+    pub share: Scalar<Secret, Zero>,
 }
 
 impl SecretShare {
@@ -63,10 +63,15 @@ impl SecretShare {
     pub fn recover_secret(shares: &[SecretShare]) -> Scalar<Secret, Zero> {
         let index_and_secret = shares
             .iter()
-            .map(|share| (share.index, share.secret))
+            .map(|share| (share.index, share.share))
             .collect::<alloc::vec::Vec<_>>();
 
         poly::scalar::interpolate_and_eval_poly_at_0(&index_and_secret[..])
+    }
+
+    /// The verification share for this secret share.
+    pub fn verification_share(&self) -> Point<NonNormal, Public, Zero> {
+        g!(self.share * G)
     }
 
     /// Encodes the secret share to 64 bytes. The first 32 is the index and the second 32 is the
@@ -74,7 +79,7 @@ impl SecretShare {
     pub fn to_bytes(&self) -> [u8; 64] {
         let mut bytes = [0u8; 64];
         bytes[..32].copy_from_slice(self.index.to_bytes().as_ref());
-        bytes[32..].copy_from_slice(self.secret.to_bytes().as_ref());
+        bytes[32..].copy_from_slice(self.share.to_bytes().as_ref());
         bytes
     }
 
@@ -83,7 +88,7 @@ impl SecretShare {
     pub fn from_bytes(bytes: [u8; 64]) -> Option<Self> {
         Some(Self {
             index: Scalar::from_slice(&bytes[..32])?,
-            secret: Scalar::from_slice(&bytes[32..])?,
+            share: Scalar::from_slice(&bytes[32..])?,
         })
     }
 }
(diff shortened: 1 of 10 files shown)


LLFourn:
This is reviewable now. Please excuse poor documentation. I think I'll do a full sweep after I've attempted implementing the bip.

See #189 for my I did multiplicative tweaking.

LLFourn:
> The awkward parts were 1) applying tweaks to both SharedKey and PairedSecretShare

Hmm I was actually thinking you wouldn't have to do this unless in your protocol signers are also verifying signature shares (which I guess in LN context is likely).

> The function signature difference between homomorphic_add and homomorphic_mul

Good point I think `homomorphic_mul` should just own self as well.

LLFourn:
@irriden I can relax this a bit but you will need `CoordinatorSignSession` to verify the signature shares anyway or are you not doing that in this case?

LLFourn:
@irriden see the two last commits. Hopefully the tweaking is more straightforward now. There is no longer two variants of tweakings "xonly" and normal tweaks. To do an xonly tweak you just do a normal `homomorphic_add` then call `non_zero` and then `into_xonly`. This is more steps for the caller but I think more natural and means less code.

I've made it so you don't need a `CoordinatorSignSession` to create the final signature. I'm guessing you are just checking the final signature for validity?

schnorr_fun/src/frost/shared_key.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// The key that was shared with this polynomial defining the sharing.
+    ///
+    /// This is the first coefficient of the polynomial.
+    pub fn key(&self) -> Point<Normal, Public, Z> {
+        Z::cast_point(self.point_polynomial[0]).expect("invariant")
+    }
+    /// Constructor to create a from a vector of points where each item represent a polynomial
+    /// coefficient.
+    ///
+    /// Returns `None` if the first coefficient is [`Point::zero`].
```
LLFourn, line 166: This is not true anymore. It can be `None` is the poly is empty. I think I might make two different `from_poly`s one for `Zero` and another for `NonZero`. The `Zero` one will be infallible.

schnorr_fun/src/frost/shared_key.rs
```diff
@@ -0,0 +1,410 @@
...
+}
+
+impl SharedKey<Normal, Zero> {
+    /// Constructor to create a shared key from a vector of points where each item represent a polynomial
+    /// coefficient.
+    ///
+    /// The resulting shared key will be `SharedKey<Normal, Zero>`. It's up to the caller to do the zero check with [`non_zero`]
+    ///
+    /// [`non_zero`]: Self::non_zero
+    pub fn from_poly(poly: Vec<Point<Normal, Public, Zero>>) -> Self {
```
nickfarrow, line 221: Is this again because there's nothing special about preventing `Zero` for a shared key?

I always feel weird seeing it allowed. Why avoid panics when the user passes an empty poly for the shared key? (maybe they just don't know it but have a secret share?)

schnorr_fun/src/frost/session.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl PartySignSession {
+    /// The final nonce that will actually appear in the signature
+    pub fn final_nonce(&self) -> Point<EvenY> {
+        self.final_nonce
+    }
+
+    /// The public key the session was started under
+    pub fn public_key(&self) -> Point<EvenY> {
```
nickfarrow, line 81: I liked how you re-allowed SharedKey to be `Normal`, but here the signing session is forced into a BIP340 context. No problem for us but maybe other people would want signing for `Normal` keys?

schnorr_fun/src/frost/share.rs
```diff
@@ -101,6 +105,158 @@ secp256kfun::impl_display_debug_serialize! {
...
+    /// Adds a scalar `tweak` to the paired secret share.
+    ///
+    /// The returned `PairedSecretShare<Normal, Zero>` represents a sharing of the original value + `tweak`.
+    ///
+    /// This is useful for deriving unhardened child frost keys from a master frost public key using
+    /// [BIP32]. In cases like this since you know that the tweak was computed from a hash of the
+    /// original key you call [`non_zero`] and unwrap the `Option` since zero is computationally
+    /// unreachable.
+    ///
+    /// If you want to apply an "x-only" tweak you need to call this then [`non_zero`] and finally [`into_xonly`].
```
nickfarrow, line 186: Makes sense, still allows for repeated x-only tweaks (forgot use-case) :+1:

schnorr_fun/src/frost/shared_key.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    /// Encodes a `SharedKey` as the compressed encoding of each underlying polynomial coefficient
+    ///
+    /// i.e. call [`Point::to_bytes`] on each coefficent starting with the constant term. Note that
+    /// even if it's a `SharedKey<EvenY>` the first coefficient (A.K.A the public key) will still be
+    /// encoded as 33 bytes.
+    ///
+    /// ⚠ Unlike other secp256kfun things this doesn't exactly match the serde/bincode
+    /// implemenations which will length prefix the list of points.
```
nickfarrow, line 170: I am curious as to why? Or is this just a note that serde/bincode prefix the length so you'll have to keep that in mind if you mix & match?

nickfarrow (approved):
I noticed you made all the public keys `shared_key` and then changed back. Looks like they are no longer going with `shared_key` on the BIP also!

Very much like 🐙 = coordinator

Proptest with all the variations of tweaks looks very robust.

One completely unrelated thing I just noticed is that `create_shares_and_pop` is kind of confusing, makes it sound like something is getting popped off a vec.

I didn't look too deeply into the secp256kfun casting between point types etc. but the idea of checking the types like this makes sense.

There's quite a few spelling mistakes, not sure whether you care much for that, happy to fix up in commit or another PR. Everything else LGTM!

schnorr_fun/src/frost/session.rs
LLFourn: There is no way to sign with `Normal` keys though.

LLFourn:
I just pushed a commit to make "verification shares" a first class type. Now you use the verification shares to verify signature shares. @irriden you can now create a polynomial from "share images" rather than "verification shares" because verification shares are strictly something you can verify a signature share against (i.e. homomorphically linked to a `Point<EvenY>`.

> One completely unrelated thing I just noticed is that `create_shares_and_pop` is kind of confusing, makes it sound like something is getting popped off a vec.

heh it will be gone soon I hope.

> There's quite a few spelling mistakes, not sure whether you care much for that, happy to fix up in commit or another PR. Everything else LGTM!

Feel free to push doc fixes you catch at anytime. I fixed up a few given that you made me self-conscious about it!

LLFourn:
Added 84ac38b593e8e6a9dcc9df1cc398d803d6eac554 which made things make a bit more sense.

PR #192, opened 2024-08-14
implement chillidkg 

LLFourn:
Implements the ideas from the bip-frost spec. This is a WIP. I think I like the new spec way better so I want to make proper errors for the error cases as well do documentation and renaming. Then delete the old key generation.

This deviates from the spec in several ways that I won't bother to list but they are mostly superficial things that I'll either fix or make an issue about.

### DIFF
--- a/schnorr_fun/src/frost/chilldkg.rs
+++ b/schnorr_fun/src/frost/chilldkg.rs
@@ -1,0 +1,1256 @@
+//! Our take on the WIP *[ChillDKG: Distributed Key Generation for FROST][ChillDKG]* spec
+//!
+//! ChillDKG is a modular distributed key generation protocol. At the end all the intended parties
+//! have a valid [Shamir secret sharing] of a secret key without the secret key ever having existed
+//! on any party's computer.
+//!
+//! The WIP spec defines two roles:
+//!
+//! - *Coordinator*: A central party who relays and aggregates messages between the other parties.
+//! - *Participants*: The parties who provide secret input and receive secret shares as output from the protocol.
+//!
+//! In this implementation we split "participants" into two further roles:
+//!
+//! - *Contributors*: parties that provide secret input into the key generation
+//! - *receivers*: parties that receive a secret share from the protocol.
+//!
+//! We see a benefit to having parties that provide secret input but do not receive secret outupt.
+//! The main example of this is having the coordinator itself be an *Contributor* too. In the context
+//! of a Bitcoin hardware wallet, the coordinator is usually the only party with access to the
+//! internet therefore, if the coordinator contributes input honestly, even if all the non-internet
+//! connected devices are malicious the *remote* adversary (who set the code of the malicious
+//! device) will not know the secret key. In fact, the adversary would have to recover `t` devics
+//! and extract their internal state to reconstruct the key. This is nice, because *in theory* and
+//! in this limited sense it gives the attacker no advantage from controlling the code of the
+//! signing devices (anyone who wants to reconstruct the key already needs `t` shares).
+//!
+//! ## Variants
+//!
+//! The spec comes in three variants:
+//!
+//! - [`simplepedpop`]: bare bones FROST key generation
+//! - [`encpedpop`]: Adds encryption to the secret input so the coordinator can aggregate encrypted secret shares.
+//! - [`certpedpop`]: `encpedpop` where each party also certifies the output so they can cryptographically convince each other that the key generation was successful.
+//!
+//! [ChillDKG]: https://github.com/BlockstreamResearch/bip-frost-dkg
+use crate::{frost::*, Schnorr};
+use alloc::{
+    collections::{BTreeMap, BTreeSet},
+    vec::Vec,
+};
+use secp256kfun::{
+    hash::{Hash32, HashAdd},
+    nonce::NonceGen,
+    poly,
+    prelude::*,
+    rand_core, KeyPair,
+};
+
+/// SimplePedPop is a bare bones secure distributed key generation aglorithm that leaves a lot left
+/// up to the application.
+///
+/// The application must figure out:
+///
+/// - How to secretly transport secret share contribution from each contributor to their intended destination
+/// - Checking that each party got the correct output by comparing [`AggKeygenInput::cert_bytes`] on each of them.
+///
+/// [`AggKeygenInput::cert_bytes`]: simplepedpop::AggKeygenInput::cert_bytes
+pub mod simplepedpop {
+    use super::*;
+    use crate::{Message, Signature};
+    use alloc::{
+        collections::{BTreeMap, BTreeSet},
+        vec::Vec,
+    };
+    use secp256kfun::hash::Hash32;
+
+    /// A party that generates secret input to the key generation. You need at least one of these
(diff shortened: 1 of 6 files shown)


LLFourn:
This is ready for review now. 

- implemented (my loose interpretation) of chilldkg
- Remove all of our existing keygen stuff
- Also fca38b59348be0c6085beeb715a7827e18f71374

schnorr_fun/src/frost/chilldkg.rs
```diff
@@ -0,0 +1,1256 @@
...
+}
+
+/// `encpedpop` is built on top of [`simplepedpop`] to add share encryption.
+///
+/// Each per recipient secret key is explicitly encrypted to each recipient and sent through the
+/// coordinator. This simplifies things a bit since all messages are to or from the coordinator. The
+/// coordinator also aggregates the ciphertexts so communication is reduced to linear in the number
+/// of participants.
+///
+/// The application still must figure out when all parties agree on the
```
nickfarrow, line 458: Unfinished comment

schnorr_fun/src/frost/chilldkg.rs
```diff
@@ -0,0 +1,1256 @@
...
+        pub fn finish(self) -> Option<AggKeygenInput> {
+            if !self.is_finished() {
+                return None;
+            }
+            let inputs = self.inputs.into_values().flatten().collect::<Vec<_>>();
+            let key_contrib = inputs
+                .iter()
+                .map(|message| (message.com[0], message.pop))
+                .collect();
+            let mut sum_coms = vec![Point::<NonNormal, Public, _>::zero(); self.threshold as usize];
```
nickfarrow, line 249: Could be worth mentioning the trick here, that the coordinator can aggregate all the coefficients except for the first coefficients, since they need to be shared so that the proof-of-possessions can be independently verified.

schnorr_fun/src/frost/chilldkg.rs
```diff
@@ -0,0 +1,1256 @@
...
+            if !check_missing.is_empty() {
+                return Err("didn't have share for all parties");
+            }
+
+            // ⚠ only do mutations after we're sure everything is ok
+            self.inner.add_input(schnorr, from, input.inner)?;
+
+            for (dest, encrypted_share_contrib) in input.encrypted_shares {
+                let agg_encrypted_share = &mut self.agg_encrypted_shares.get_mut(&dest).unwrap().1;
+                *agg_encrypted_share += encrypted_share_contrib;
```
nickfarrow, line 751: I was going to suggest doing this aggregation in `finish` but actually it's good here since you can reuse simplepedpop types nicely

schnorr_fun/src/frost/chilldkg.rs
```diff
@@ -0,0 +1,1256 @@
...
+        encryption_jobs: BTreeMap<PartyIndex, (Point, Scalar<Secret, Zero>)>,
+        multi_nonce_keypair: KeyPair<Normal>,
+    ) -> BTreeMap<PartyIndex, Scalar<Public, Zero>> {
+        encryption_jobs
+            .iter()
+            .map(|(dest, (encryption_key, share))| {
+                let dh_key = g!(multi_nonce_keypair.secret_key() * encryption_key).normalize();
+                let pad = Scalar::from_hash(H::default().add(dh_key).add(encryption_key).add(dest));
+                let payload = s!(pad + share).public();
+                (*dest, payload)
```
nickfarrow, line 827: This doesn't quite match the encryption of `encpedpop` [here](https://github.com/BlockstreamResearch/bip-frost-dkg/blob/master/python/chilldkg_ref/encpedpop.py#L17).

We calulate dh_key the same way.
Their hash is different to ours:
- `Hash( "encpedpop ecdh", dh_key ||  my_pubkey || their_pubkey || context_data )`
- I think their `my_pubkey`/`pubnonce` we have as `multi_nonce_keypair.public_key()`, not `encryption_key`?
- Our hash is missing context data that sounds important? [See this comment:](https://github.com/BlockstreamResearch/bip-frost-dkg/blob/master/python/chilldkg_ref/encpedpop.py#L40) This hashed ElGamal's "crucial feature is to feed the index of the enckey to the hash function"

secp256kfun/src/hash.rs
```diff
@@ -150,6 +150,15 @@ impl<K: HashInto, V: HashInto> HashInto for alloc::collections::BTreeMap<K, V> {
     }
 }
 
+#[cfg(feature = "alloc")]
+impl<T: HashInto + Ord> HashInto for alloc::collections::BTreeSet<T> {
+    fn hash_into(self, hash: &mut impl digest::Update) {
+        for item in self {
```
nickfarrow, line 156: Will this item ordering be consistent with other implementations?
Maybe a slice of party indices would be better?

schnorr_fun/src/frost/chilldkg.rs
LLFourn: I instead implemented it like in the paper referenced:

image

1. Yes there a few instances of this. I am just putting in what I think is actually important.
2. `my_pubkey` is  `encryption_key` and `pubnonce` is `multi_nonce_keypair.public_key()`. I'm not sure what the right name. Maybe it shouldn't be called encryption_key because we later transform it into a signing key.
3. The index is dest.

secp256kfun/src/hash.rs
LLFourn: Consistent with other implementations of hashing BTreeSets? I mean there's only one way to hash an ordered set...
With respect to FROST signing I don't see any reason why anyone would order things any differently or why we should worry about that now.

schnorr_fun/src/frost/chilldkg.rs
LLFourn: I made an explanation of why the first coefficients aren't aggregated.

schnorr_fun/src/frost/chilldkg.rs
LLFourn: The simpler state is better.

schnorr_fun/src/frost/chilldkg.rs
nickfarrow: That makes sense. I also misread their code, they are computing the hash in the same way.

secp256kfun/src/hash.rs
nickfarrow: I was more-so thinking for future compatibility with other implementations that might not use BTreeSets (python). This ordering on each signer will need to match to produce a valid signature. But I forgot that BTreeSet uses the underlying Ord, so the underlying sorting here likely already makes sense.

nickfarrow (approved):

schnorr_fun/src/frost/chilldkg.rs
LLFourn: Yes. `pubnonce` is definitely superfluous since it's a function of `dh_key` and `my_pubkey`. There is an approach to these kinds of "what to hash" question which is just hash every single thing without giving a particular justification for it. That's a fine approach but it requires extra care when specifying the thing otherwise you end up with concatenating stuff like "context_data" everywhere and it gets unwieldy.

PR #206, opened 2025-02-09
chore: Nonce::from_bytes() returns Self instead of Option<Self>

tvolk131:
`Point::from_slice()` only returns `None` if it is passed a byte slice that isn't exactly 33 bytes long. Since `Nonce::from_bytes()` always passes in 33 bytes, we can guarantee to the caller that this never fails.


nickfarrow:
Hey @tvolk131, I asked Lloyd about this one and it due to the fact that not every 33 byte array will render a valid Point, see `Point::from_bytes`. For example, the first byte must be either `0x02` or  `0x03` to indicate that the y-coordinate is even/odd.

LLFourn:
Yes it's very important that not all 33 byte arrays are valid points because not all 32 byte arrays are valid x-coordinates on secp256k1 (around half are not!).

PR #210, opened 2025-06-13
Upgrade to bincode v2 actually released on crates.io

LLFourn:


LLFourn:
This also updates the MSRV since the released version of `bincode_derive` needs 1.85.0

PR #217, opened 2025-06-19
Implement hash_to_curve for secp256k1

LLFourn:
## Summary

Implements hash-to-curve functionality for secp256k1 following the IETF draft specification:
https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/

This was necessary because the upstream k256 implementation couldn't be directly copied into secp256kfun due to dependency and API differences.

## Changes

- Added `hash_to_curve` module in vendor directory
- Exposed through `BackendPoint` trait 
- Added public API on `Point` that returns `NonNormal` points
- Made `pow2k` method `pub(crate)` in vendor/k256/field.rs to reuse it
- Comprehensive tests including RFC test vectors and property-based tests

<details>
<summary>Conversation History</summary>

**User**: implement hash_to_curve using https://github.com/RustCrypto/elliptic-curves/blob/master/k256/src/arithmetic/hash2curve.rs as a guide for the implementation. It should pass the exact same test vectors. It should use the Hash32 trait as a type parameter for the hash function. It should use sha2 in the tests. it should be using the vendor module. Do not expose the vendor types outside of the vendor module. First create an API skeleton with todo\!() and complete test vectors. Make sure it all compiles. Then implement it.

**Assistant**: [Created initial skeleton with test vectors and API structure]

**User**: You can just condition the whole test block on `alloc`. std implies alloc. I want you to implement the test vectors completely. They won't pass of course but they should be fully there. Don't worry about writing the function signatures for the API if you're not implementing it just leave the body as empty or todo\!() for now. Use the hex.rs decode fns to decode the test vectros.

**Assistant**: [Implemented complete test vectors]

**User**: Can we make a sort of dummy `hex\!` macro here that just does decode_array and unwrap. to clean up the test vectors.

**Assistant**: [Created hex\! macro and cleaned up test vectors]

**User**: ok great now try and implement hash_to_curve exactly like in elliptic-curves k256 and pass the tests\!

**Assistant**: [Implemented hash_to_curve following k256]

**User**: the hash to curve impls shouldn't need alloc. All the sizes should be known up front. Just make arrays.

**Assistant**: [Refactored to use fixed-size arrays instead of Vec]

**User**: on top of that why are you directly using the particular backend field element type like `FieldElement5x52`. You should only be using the public `FieldElement` type we have re-exported. You seem to have avoided using FieldElement::from_bytes_unchecked for all these little field element constants. This is a mistake. This is a const function you can use to declare all these constants in a similar manner to the k256 original implementation.

**Assistant**: [Fixed to use public FieldElement API and from_bytes_unchecked for constants]

**User**: [After debugging] Did you realize that pow2k is a method on FieldElement already you didn't have to define it?

**User**: well it's not part of the public api but I wouldn't mind if we made it pub(super) or something like that.

**Assistant**: [Made pow2k pub(crate) and removed duplicate implementation]

**User**: I think the same thing with 48? This is fixed in the spec right? If so let's just make a const for it and don't pass it in as a parameter anywhere.

**Assistant**: [Made L a constant = 48]

**User**: it seems like from_okm can take a fixed length array now

**Assistant**: [Updated from_okm to take &[u8; L]]

**User**: can you explain a bit why we need to know the BlockSize of the hash function for this. Why does the spec care?

**Assistant**: [Explained Z_pad security padding in expand_message_xmd]

**User**: ok let's get rid of the OutputSizeUser thing at least. It's Hash32 which has a fixed output size.

**Assistant**: [Removed OutputSizeUser trait bound]

**User**: You have these comments like `// Compute x_num = XNUM[0] + XNUM[1] * x + XNUM[2] * x^2 + XNUM[3] * x^3` why not just define the constants as arrays so you could just refer to them as XNUM[0] etc.

**Assistant**: [Refactored isogeny constants to arrays]

**User**: These comments seem very redundant now. Can you not have a comment above the line of code saying almost the exact same thing.

**Assistant**: [Removed redundant comments]

**User**: map_to_curve doesn't make sense. It for some reason turns the return value into Affine and then returns Projective. why don't you just return an AffinePoint directly.

**Assistant**: [Fixed map_to_curve to return AffinePoint]

**User**: please remove the clear cofactor stuff.

**Assistant**: [Removed clear_cofactor as secp256k1 has cofactor=1]

**User**: Ok so what we're going to do is expose this hash_to_curve on the BackendPoint trait and impl it using what you've done.

**Assistant**: [Added hash_to_curve to BackendPoint trait and implemented it]

**User**: Err I don't think that hash_to_curve returns a Normal point. It should be NonNormal. It comes from a projective point. Don't normalize it aggressively.

**Assistant**: [Fixed to return NonNormal point]

**User**: nooo you shouldn't use the internals in the point.rs tests. Ok to make sure you don't make this mistake put the tests in tests/test_hash_to_curve.rs so they can only use the public API.

**Assistant**: [Moved tests to integration tests using only public API]

**User**: can you turn the test_hash_to_curve_public_api thing into a proptest with different random strings.

**Assistant**: [Added proptest for property-based testing]

**User**: proptest is a dev dependency it doesn't need a feature gate in test files

**Assistant**: [Removed unnecessary feature gates]

**User**: In `expand_message_xmd` the final parameter is an invariant.

**Assistant**: [Made len_in_bytes parameter an invariant]

**User**: ell is also an invariant now

**Assistant**: [Made ell an invariant]

**User**: ok I feel like I've made a mess of this. I think you should return it to the more spec comparable version which took the argument and implemented this whole crazy construction. I thought it would get less crazy as you take care of the invariants but it doesn't at all.

**Assistant**: [Reverted expand_message_xmd to spec-compliant version with parameters]

</details>

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/secp256kfun/src/point.rs
+++ b/secp256kfun/src/point.rs
@@ -1,7 +1,7 @@
 use crate::{
     Scalar,
     backend::{self, BackendPoint, TimeSensitive},
-    hash::HashInto,
+    hash::{Hash32, HashInto},
     marker::*,
     op,
 };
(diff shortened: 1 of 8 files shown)


secp256kfun/src/vendor/hash_to_curve.rs
```diff
@@ -0,0 +1,568 @@
...
+    let gxd = tv2 * xd;
+    tv2 = tv2 * MAP_A;
+
+    let mut gx1 = x1n * (tv2 + x1n.square());
+    tv2 = gxd * MAP_B;
+    gx1 = gx1 + tv2;
+
+    let mut tv4 = gxd.square();
+    tv2 = gx1 * gxd;
+    tv4 = tv4 * tv2;
```
LLFourn, line 313: I'm pretty sure you can do: `tv4 *= tv2`. There's a few of places where this operator could be used to simplify the code in this function.

secp256kfun/src/point.rs
```diff
@@ -114,6 +114,175 @@ impl Point<Normal, Public, NonZero> {
...
+    /// let point = Point::hash_to_curve_rfc9381_tai::<Sha256>(b"hello world", b"my-salt");
+    /// // Use empty bytes if no salt is needed
+    /// let point2 = Point::hash_to_curve_rfc9381_tai::<Sha256>(b"hello world", b"");
+    /// ```
+    ///
+    /// [`hash_to_curve_sswu`]: Self::hash_to_curve_sswu
+    pub fn hash_to_curve_rfc9381_tai<H: Hash32>(
+        msg: &[u8],
+        salt: &[u8],
+    ) -> Point<Normal, Public, NonZero> {
```
LLFourn, line 251: This should return an `EvenY` point if what the documentation says is correct. It should use `from_xonly_bytes.`

secp256kfun/src/point.rs
```diff
@@ -114,6 +114,173 @@ impl Point<Normal, Public, NonZero> {
...
+    /// [`hash_to_curve_sswu`]: Self::hash_to_curve_sswu
+    /// [RFC 9381]: https://datatracker.ietf.org/doc/html/rfc9381
+    pub fn hash_to_curve<H: Hash32>(hasher: H) -> Point<Normal, Public, NonZero> {
+        use crate::hash::HashAdd;
+
+        // Try up to 255 times (probability of failure is negligible)
+        for counter in 0u8..u8::MAX {
+            let hash_bytes = hasher.clone().add(counter).finalize_fixed();
+
+            // Use 0x03 (odd y) when counter==0, 0x02 (even y) when counter>0
```
LLFourn, line 207: This comment is wrong it's 0x02 when counter == 0.

secp256kfun/src/point.rs
```diff
@@ -114,6 +114,174 @@ impl Point<Normal, Public, NonZero> {
...
+    /// runtime based on input (see [`hash_to_curve_sswu`] for details), this is rarely
+    /// a practical concern.
+    ///
+    /// ## Why not the [RFC 9381] try-and-increment?
+    ///
+    /// The VRF specification ([RFC 9381 §5.4.1.1](https://datatracker.ietf.org/doc/html/rfc9381#section-5.4.1.1))
+    /// includes a try-and-increment method (see [`hash_to_curve_rfc9381_tai`]) that always
+    /// uses a fixed y-coordinate parity (0x02). This results in a non-uniform distribution
+    /// that only includes points with even y-coordinates. Our implementation achieves
+    /// uniform distribution with a simple modification.
```
nickfarrow, line 187: Are there any consequences of only outputting even-y hash to curve points? If not, i'm thinking that using and recommending this RFC could be best and most widely supported? Even if the change is minor

PR #222, opened 2025-06-23
Make certpedpop signature scheme configurable

LLFourn:
## Summary

This PR refactors the `certpedpop` module to make the signature scheme configurable by applications, rather than being hardcoded to Schnorr signatures. This allows users to plug in their preferred signature scheme (ECDSA, EdDSA, etc.) while maintaining backward compatibility with the existing Schnorr implementation.

## Key Changes

- **Added `CertificationScheme` trait**: Defines the interface for signature schemes that can certify DKG output
- **Generic implementation**: Made `Certificate`, `CertifiedKeygen`, and related functions generic over the certification scheme
- **Schnorr implementation**: Provided a default implementation of `CertificationScheme` for the existing Schnorr signature scheme (BIP340)
- **Simplified key types**: Changed certification keys from `Point<EvenY>` to `Point` for greater flexibility

## API Changes

The main API change is that functions now require a certification scheme parameter:

```rust
// Before
let sig = contributor.verify_agg_input(&schnorr, &agg_input, &keypair)?;

// After  
let sig = contributor.verify_agg_input(&cert_scheme, &agg_input, &keypair)?;
```

For backward compatibility, users can simply pass their Schnorr instance as both the regular parameter and the certification scheme.

## Test Plan

- [x] All existing tests pass
- [x] `cargo check --all-features` passes
- [x] `cargo clippy --all-features` shows no warnings
- [x] `cargo fmt` has been run

<details>
<summary>Conversation History</summary>

**User**: Ok in chilldkg.rs I want certpedpop to let the application to decide what kind of signature scheme it wants to use for the certification.

**Assistant**: I'll help you modify the certpedpop implementation in chilldkg.rs to make the signature scheme configurable by the application.

**User**: I think we have an invariant that the PublicKey has to be Point (since it is also used as an encryption key elsewhere). I think the message should be typed with a AggkeygenInput since that's what we're certifying.

**Assistant**: You're absolutely right\! Let me revise the trait design based on those constraints...

**User**: I'm not sure it should be Point<EvenY> actually maybe just Point. I know you'll have to change it in the Certificate thing as well. Also don't have an error types. Just return bool for verify.

**Assistant**: You're right\! Here's the updated design...

**User**: Given the public key is fixed I think the "SigningKey" should just be a `KeyPair`.

**Assistant**: You're absolutely right\! Since the public key is `Point`, the signing key should be `KeyPair` to maintain consistency.

**User**: ok define this trait in the file so we don't lose it

**Assistant**: [Added the CertificationScheme trait definition to the file]

**User**: Make an implementation of the trait using standard Schnorr signatures from this crate (BIP340)

**Assistant**: [Implemented CertificationScheme for Schnorr with BIP340 signatures]

**User**: alright have a shot at integrating it into certpedpop

**Assistant**: [Refactored the entire certpedpop module to use the generic CertificationScheme trait]

</details>

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/schnorr_fun/src/frost/chilldkg.rs
+++ b/schnorr_fun/src/frost/chilldkg.rs
@@ -47,6 +47,50 @@ use secp256kfun::{
     rand_core,
 };
 
+/// A trait for signature schemes that can be used to certify the DKG output.
+///
+/// This allows applications to choose their preferred signature scheme for
+/// certifying the aggregated keygen input in certpedpop.
+pub trait CertificationScheme {
+    /// The signature type produced by this scheme
+    type Signature: Clone + core::fmt::Debug + PartialEq;
+
+    /// Sign the AggKeygenInput with the given keypair
+    fn certify(&self, keypair: &KeyPair, agg_input: &encpedpop::AggKeygenInput) -> Self::Signature;
+
+    /// Verify a certification signature
+    fn verify_cert(
+        &self,
+        cert_key: Point,
+        agg_input: &encpedpop::AggKeygenInput,
+        signature: &Self::Signature,
+    ) -> bool;
+}
+
+/// Standard Schnorr (BIP340) implementation of the CertificationScheme trait
+impl<H: Hash32, NG: NonceGen> CertificationScheme for Schnorr<H, NG> {
+    type Signature = crate::Signature;
+
+    fn certify(&self, keypair: &KeyPair, agg_input: &encpedpop::AggKeygenInput) -> Self::Signature {
+        let cert_bytes = agg_input.cert_bytes();
+        let message = crate::Message::<Public>::plain("BIP DKG/cert", cert_bytes.as_ref());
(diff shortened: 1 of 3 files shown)


schnorr_fun/src/frost/chilldkg.rs
```diff
@@ -1019,19 +1063,21 @@ pub mod certpedpop {
         /// can use the share you must call [`finalize`] with a completed certificate.
         ///
         /// [`finalize`]: Self::finalize
-        pub fn receive_share<H, NG>(
+        pub fn receive_share<H, NG, S>(
```
LLFourn, line 1066: We should rename this to receive_secret_share and call this SecretShareReceiver.

nickfarrow (approved):
Noticed the branch name `generic-bincode-challenge-length` doesn't match what this PR covers, is there more to come? Otherwise LGTM

PR #226, opened 2025-08-13
feat: Add ShareImage type for frost share images

LLFourn:
## Summary

Introduced a dedicated `ShareImage` struct to encapsulate the public information about a secret share (index and point). This provides a cleaner API for working with share images, particularly for reconstructing shared keys from threshold share images.

<details><summary>Conversation History</summary>

User: ok can you find out of there's a `ShareImage` concept anywhere?
Assistant: [Searched and found no ShareImage in the codebase]

User: ok can you look at ~/src/frostsnap2 and look at how ShareImage works?
Assistant: [Found ShareImage in frostsnap2 - a struct with index: PartyIndex and image: Point]

User: cool I want you to port the concept over to schnorr_fun. Put it in the other file where there's share stuff like "PairedSecretShare".
Assistant: [Added ShareImage struct to schnorr_fun/src/frost/share.rs]

User: ShareImage shouldn't have type parameters I don't think. The point should be Normal, Public, Zero for now. I don't think you'll need the serde and bincode bounds there. You can just use the feature gated plain derives.
Assistant: [Updated ShareImage to remove type parameters and simplified derives]

User: I think there's also a method on SharedKey that could benefit from this type which mentions it uses "share images".
Assistant: [Updated SharedKey methods to use ShareImage type]

User: In the docs for ShareImage can you mention and link to the SharedKey function that can recompute the SharedKey from the images.
Assistant: [Added documentation link to SharedKey::from_share_images]

</details>

### DIFF
--- a/schnorr_fun/src/frost/share.rs
+++ b/schnorr_fun/src/frost/share.rs
@@ -1,3 +1,4 @@
+use super::ShareIndex;
 use secp256kfun::{poly, prelude::*};
 /// A *[Shamir secret share]*.
 ///
@@ -87,8 +88,11 @@ impl SecretShare {
     }
 
     /// Get the image of the secret share.
-    pub fn share_image(&self) -> Point<NonNormal, Public, Zero> {
-        g!(self.share * G)
+    pub fn share_image(&self) -> ShareImage {
+        ShareImage {
+            index: self.index,
+            image: g!(self.share * G).normalize(),
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


schnorr_fun/src/frost/share.rs
```diff
@@ -501,7 +505,27 @@ mod share_backup {
 #[cfg(feature = "share_backup")]
 pub use share_backup::BackupDecodeError;
 
-use super::ShareIndex;
+/// The public image of a secret share, consisting of an index and the corresponding point.
+///
+/// A `ShareImage` represents the public information about a share: the index at
+/// which the polynomial was evaluated and the image of the secret share. This
+/// can be shared publicly and used to reconstruct the shared public key from a
```
LLFourn, line 512: should be shared public key *(and the polynomial)*.

nickfarrow (approved):

PR #230, opened 2025-08-26
[❄] Add SharedKey::grind_fingerprint

LLFourn:
So we can do fingerprint grinding outside of chilldkg module.

### DIFF
--- a/schnorr_fun/src/frost/chilldkg/encpedpop.rs
+++ b/schnorr_fun/src/frost/chilldkg/encpedpop.rs
@@ -216,82 +216,45 @@ impl AggKeygenInput {
             .ok_or("the shared secret was zero")
     }
 
-    /// Grinds all polynomial coefficients to achieve a fingerprint by rejection sampling.
-    /// This is meant to be run by the coordinator.
+    /// Embeds a proof-of-work `fingerprint` into the aggregated polynomial.
     ///
-    /// This method modifies each non-constant coefficient of the polynomial through group
-    /// addition until the running hash (computed by sequentially incorporating coefficients)
-    /// has the required number of leading zero bits at each step.
+    /// This coordinator-only operation modifies the DKG output to include
+    /// a verifiable proof of work by grinding the polynomial coefficients.
+    /// The `fingerprint` specifies the required difficulty (number of leading
+    /// zero bits) and an optional tag to include in the hash.
     ///
-    /// The fingerprint is tied to the specific public key by including it in the hash.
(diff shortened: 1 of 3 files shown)


schnorr_fun/src/frost/chilldkg/encpedpop.rs
```diff
@@ -216,82 +216,45 @@ impl AggKeygenInput {
...
+        self.inner.agg_poly = shared_key.point_polynomial()[1..].to_vec();
+        debug_assert!(self.shared_key().check_fingerprint::<H>(&fingerprint));
+
+        for (share_index, (_encryption_key, encrypted_secret_share)) in &mut self.encrypted_shares {
+            // 💡 The share encryption is homomorphic so we can apply the tweak
+            // operations as if the coordinator had local copies of the secret
+            // shares.
+            let mut tmp = SecretShare {
+                index: *share_index,
+                share: encrypted_secret_share.secret(/*HACK: we need a secret one to use the function*/),
```
nickfarrow, line 254: `encrypted_secret_share.secret()` was quite confusing since it sounds like we're getting the secret out of the encrypted secret share, not *marking* it as secret for the API.

Checking understanding:

We find grind some tweaks to the non-constant terms of the polynomial.
Then we overwrite the non-constant terms of `agg_poly` from the grinded polynomial.

Then finally, we modify all the encrypted secret share by adding the term to have it sit on this tweaked polynomial?

LLFourn: yes. I made it easier to read by having `SecretShare` take a type parameter for the secrecy so you don't have to cast it.

schnorr_fun/src/frost/shared_key.rs
```diff
@@ -186,35 +186,32 @@ impl<T: PointType, Z: ZeroChoice> SharedKey<T, Z> {
...
     pub fn check_fingerprint<H: crate::fun::hash::Hash32>(
         &self,
         fingerprint: &Fingerprint,
     ) -> bool {
         use crate::fun::hash::HashAdd;
 
-        if self.point_polynomial.is_empty() {
+        // the fingerprint is only placed on the non-constant coefficients so it
+        // can't be detected with a length 1 polynomial
+        if self.point_polynomial.len() <= 1 {
```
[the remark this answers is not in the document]
LLFourn: Yes we consider the fingerprint to always be correct for it.

nickfarrow (approved):

PR #232, opened 2025-08-27
No sha2 dep

LLFourn:
on top of #231 

removes the sha2 dependency from schnorr_fun slightly changing the API to compensate.


vrf_fun/tests/codec.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "bincode")]
```
nickfarrow, line 1: Redundant? We already feature gate the tests in this file

LLFourn: no because of warnings. Nothing in this file does anything without that feature so it should be excluded without it.

nickfarrow (dismissed):

nickfarrow (approved):

nickfarrow (dismissed):

nickfarrow (approved):

PR #238, opened 2025-09-05
Add TryFrom integer conversions for NonZero scalars

LLFourn:
Adds  that fails when the input is zero, with a proper  error type.


secp256kfun/src/scalar.rs
```diff
@@ -497,6 +521,27 @@ mod conversion_impls {
...
+
+                    fn try_from(value: $t) -> Result<Self, Self::Error> {
+                        // big-endian integer → 32-byte array
+                        let mut bytes = [0u8; 32];
+                        let int_bytes = value.to_be_bytes();
+                        bytes[32 - int_bytes.len() ..].copy_from_slice(&int_bytes);
+                        let scalar = Scalar::<S, Zero>::from_bytes(bytes).unwrap();
+
+                        // Check if value is zero
+                        if value == 0 {
```
nickfarrow: nit: could check this first for an early return before doing the computation. Maybe negligible, and maybe the compiler would optimize this

nickfarrow (approved):

PR #247, opened 2026-04-21
chilldkg: security review fixes

### LINKED ISSUE #1: Shouldn't messages to be signed be marked with a secrecy?
When we produce the Fiat-Shamir challenge for a Schnorr signature we return a scalar that is `<Public, Zero>`: https://github.com/LLFourn/secp256kfun/blob/6dd73322af7c57f98b3217342b6dfdfa004e77a3/schnorr_fun/src/lib.rs#L97-L108

But what if the message is secret? It is possible that operations on this scalar will be done in variable leaking decisional information about the message.

I think the best decision here is to have a `Message<&'a, S: Secrecy>` struct that holds  a reference to a message slice. Perhaps we can accept anything that `Into<Message>`. By default `&[u8]` can be `Public` but at least you can opt in to strict constant time verification type algorithms.

[EDIT] This is not so easy. It needs to be that if anything secret goes into the hash then the output should be marked secret. But these rules are hard to create in rust at the moment. I think this requires higher kinded types.

LLFourn:
## Summary

A security review of the `chilldkg` module in `schnorr_fun` (and adjacent code reached from its decoding paths). **No critical vulnerabilities were found.** The issues fall into two buckets:

1. **Decoding panics** — cases where malformed serialized input could crash the process rather than return an error.
2. **Defensive hardening** — places where the protocol was permissive in ways that didn't expose secrets but tolerated unintended behaviour (ghost contributions, x-only PoP replay, silently accepted unverified certs, redundant/confusing parameters, etc.).

Each commit is self-contained and independently reviewable.

## Commit-by-commit

**`b41919d` — Replace &'static str errors in chilldkg with typed error enums**

Introduces `simplepedpop::AddInputError`, `encpedpop::AddInputError`, `encpedpop::RecoverShareError`, `certpedpop::RecoverShareError`. Replaces string-typed `Result<_, &'static str>` returns so callers can discriminate failures programmatically. Drops the unused `CertifierError::DuplicateCertificate` variant.

**`58d90d9` — Fix panics on malformed input in chilldkg and SharedKey decoding**

Three panic paths reachable from untrusted input:

- `encpedpop::Contributor::verify_agg_input` indexed `encryption_nonces[my_index]` directly. A coordinator sending a shorter vec crashed the contributor; now returns `ContributionDidntMatch`.
- `SharedKey::from_slice(&[])` and the `bincode::Decode` / `serde::Deserialize` impls panicked on empty polynomials via `poly[0]`. Now return `None` / a decode error.
- `simplepedpop::Contributor::gen_keygen_input` indexed `secret_poly[0]` when called with `threshold == 0`. Now asserts `threshold > 0` matching `Coordinator::new`.

**`b638920` — Enforce contributor count across chilldkg layers**

A malicious coordinator can craft an `AggKeygenInput` with more key_contrib slots than any honest contributor agreed to (a "ghost" contribution whose secret the coordinator knows). Doesn't compromise secrecy but gives the coordinator extra influence over the public key and breaks the protocol's implicit "every contributor signed off" contract.

- `simplepedpop::Contributor` now stores `n_contributors` at construction and `verify_agg_input` rejects mismatched `key_contrib.len()`.
- `encpedpop::Contributor::verify_agg_input` additionally rejects mismatched `encryption_nonces.len()`.
- `certpedpop::Certifier::new` becomes fallible and rejects an `AggKeygenInput` whose contributor count doesn't match `contributor_keys.len()`, with new variant `CertifierError::ContributorCountMismatch`.

**`4ff15a0` — Bind simplepedpop proof-of-possession to its slot index and parity**

The PoP previously signed the empty message. Concrete replay primitives this allowed:

- `(com, pop)` from one slot replayed at a different slot.
- `(−com, pop)` accepted under BIP340 x-only verification even though the caller intended `com`.

The PoP message is now `parity_byte || idx_be` (5 bytes), carried under the BIP340-style 33-byte domain separator `"BIP DKG/pop"`. Parity binding closes the x-only gap — `(−A, pop_A)` reconstructs a different message and fails verify. This diverges from the bip-frost-dkg Python reference's choice to bind only `idx` via BIP340's `tag_prefix`; we bind more and use the BIP340-recommended message-prefix domain separator. Regression test covers cross-slot replay and negated-key-same-slot replay.

**`21ce300` — Remove serialization from CertifiedKeygen**

The only sound path to a `CertifiedKeygen` is `Certifier::finish`, where every stored certificate was verified as it came in. Deserialization bypassed that and could land attacker-chosen gammas in the struct; `vrf_security_check` would then hand back a beacon value the attacker controlled. Dropping the `bincode::Encode`/`Decode` and serde `Serialize`/`Deserialize` derives on `CertifiedKeygen` eliminates the unverified-state path. Apps that need to persist a certified keygen can serialize `AggKeygenInput` and the certificate map separately and rebuild on the receiving side.

**`b1e1c4e` — Validate share index and encryption key in receive_secret_share**

`encpedpop::receive_secret_share` silently tolerated two malformed inputs: a missing entry in `encrypted_shares` (treated as a zero ciphertext via `unwrap_or_default`) and a caller-supplied keypair that didn't match the registered encryption key. Both surfaced as the generic `InvalidSecretShare`, diverging from the sibling `AggKeygenInput::recover_share` path which already returned `UnknownShareIndex` / `WrongEncryptionKey`. Adds both variants to `ReceiveShareError` and checks them explicitly before decryption.

**`9c34719` — Always verify signatures in Certifier::receive_certificate**

On a duplicate cert (same key already registered), the function returned `Ok` without calling `verify_cert` on the new signature. A `debug_assert_eq!` guarded the "sigs must match" case but vanished in release builds, so a mismatched second signature was silently discarded while the caller believed the cert was validated. For randomized signature / VRF schemes this was the always-taken path. Now every supplied signature is verified; a verified duplicate is idempotent (first stored wins), a failed verification returns `InvalidSignature`.

<details><summary>Conversation History</summary>

**User:** can you check for security issues in chilldkg's implementation in schnorr_fun. I am especially concerned with parties not checking their inputs and included in things.

**Assistant:** (Flagged 7 concerns: panic in encpedpop `verify_agg_input`; ghost contributor injection; `receive_secret_share` keypair validation; PoP not bound to context; Certifier silent duplicate-cert; `vrf_security_check` trusts unverified gammas; same-keypair-for-encryption-and-cert design note.)

**User:** cool fix #1 and then we'll discuss #2.

**User:** actually before moving on to #2 can you search for similar problems especially in decoding implementations that could cause panics.

**Assistant:** (Found `SharedKey::from_slice(&[])` panic, bincode Decode / serde Deserialize empty-poly panics, `gen_keygen_input` threshold=0 panic.)

**User:** Do you think that #2 is a security vulnerability? Can you explain the attack.

**Assistant:** (Ghost injection: count defense missing means malicious coordinator can add slots with known-secret "ghost" contributions. Not a secrecy break but adaptive-bias on the public key; protocol doesn't enforce "N authorized contributors.")

**User:** right but the thing is that in practice the coordinator is always a contributor as well... But I agree it's sort of worth fixing since technically the coordinator role itself is not meant to have that ability.

**User:** why couldn't the attacker just echo back the signature shares contributions of the honest party too (except negated). They are homomorphic.

**Assistant:** (Walked through why the negation attack doesn't close into key compromise: `encpedpop` pads are `H(dh_key, enc_key, dest)` hash outputs, not linearly composable. The attack remains a targeted DoS; encryption-layer non-linearity keeps the downstream damage bounded.)

**User:** Ok can you look up the python referenmce for the FROST chilldkg and see what they throw in the PoP and we can see if we can follow it.

(— bip-frost-dkg signs `idx.to_bytes(4, big)` under tag `"BIP DKG/pop message"`. schnorr_fun uses BIP340 message-prefix domain separation instead of `tag_prefix`. Bound `idx` here.)

**User:** @simplepedpop.rs I feel like [the x-only same-slot replay passing] is not a very desirable behaviour.

(— the test acknowledged that negated-key-same-slot passed PoP and was only caught by `verify_agg_input`'s Point-equality self-check. User wanted the PoP itself to bind parity. Extended the PoP message to `parity_byte || idx_be`.)

**User:** maybe we just shouldn't allow [CertifiedKeygen] to be serialized?

(— chose the simplest fix for `vrf_security_check`'s reliance on unverified gammas: drop the derive.)

**User:** A just drop it [the n_contributors param from receive_secret_share]. When you're receiving a secret share the only thing you care about is it's the correct secret share with respect to the join sharing polynomial.

(— decided `receive_secret_share`'s `n_contributors` parameter was redundant with `agg_input`; receivers' defense against count tampering is the cert round or out-of-band comparison, not a per-receive assertion. Folded into `b638920`.)

**User:** What is the actual security implications of duplicate keys though in general... I am not really interested in adding more errors I am more interested in avoiding them when the protocol could tolerate duplicates.

(— audited for panics / spurious errors on duplicate keys across `Certifier::required_keys`, `certificates`, and related paths; found none; concluded the `BTreeSet` dedup is intentional and benign. No code changes.)

</details>

### DIFF
--- a/schnorr_fun/src/frost/chilldkg/certpedpop.rs
+++ b/schnorr_fun/src/frost/chilldkg/certpedpop.rs
@@ -95,6 +95,12 @@ impl Contributor {
     /// This method is used when a party acts as both a contributor (providing entropy)
     /// and a receiver (getting a secret share), using the same keypair for both
     /// encryption/decryption and certification.
+    ///
+    /// Returns the same shape as the receiver-only entry point
+    /// [`SecretShareReceiver::receive_secret_share`]: a [`SecretShareReceiver`]
+    /// that *holds* the share until [`SecretShareReceiver::finalize`] is
+    /// called with a complete cert map. Callers must not use the share before
+    /// finalize succeeds — otherwise they're trusting an uncertified keygen.
     pub fn verify_receive_share_and_certify<H: Hash32, NG: NonceGen, S: CertificationScheme>(
         self,
         pop_schnorr: &Schnorr<H, NG>,
@@ -102,7 +108,7 @@ impl Contributor {
         share_index: ShareIndex,
         keypair: &KeyPair,
         agg_input: &AggKeygenInput,
-    ) -> Result<(PairedSecretShare<Normal, Zero>, S::Signature), CombinedRoleError> {
+    ) -> Result<(SecretShareReceiver, S::Signature), CombinedRoleError> {
         // First verify my contribution was included
         self.inner
             .verify_agg_input(agg_input)
@@ -116,7 +122,13 @@ impl Contributor {
         // Finally certify the result
         let sig = cert_scheme.certify(keypair, agg_input);
 
-        Ok((paired_secret_share, sig))
+        Ok((
+            SecretShareReceiver {
+                paired_secret_share,
+                agg_input: agg_input.clone(),
+            },
+            sig,
+        ))
     }
 }
 
@@ -313,7 +325,9 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
     )
     .expect("simulate_keygen produces matching contributor counts");
 
-    let mut paired_secret_shares = vec![];
+    // Hold dual-role receiver state until certification finishes; only then
+    // can we release the paired share via finalize.
+    let mut dual_role_receivers: Vec<(SecretShareReceiver, KeyPair)> = vec![];
 
     // Handle parties that are both contributors and receivers using the combined API
     for (i, (party_index, enckey)) in receiver_enckeys
@@ -322,7 +336,7 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
         .take(n_receivers as usize)
     {
         // This party is both a contributor and receiver - use combined method
-        let (paired_secret_share, sig) = contributors[i]
+        let (receiver, sig) = contributors[i]
             .clone()
             .verify_receive_share_and_certify(
                 schnorr,
@@ -338,7 +352,7 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
             .receive_certificate(enckey.public_key(), sig)
             .unwrap();
 
-        paired_secret_shares.push((paired_secret_share.non_zero().unwrap(), *enckey));
+        dual_role_receivers.push((receiver, *enckey));
     }
 
     // Handle extra contributors that are only contributors (not receivers)
@@ -356,6 +370,18 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
     let certified_keygen = certifier
         .finish()
         .expect("Certifier should have all required certificates");
+
+    // Now that certification is done, release each dual-role party's share.
+    let cert_map = certified_keygen.certificate().clone();
+    let paired_secret_shares: Vec<(PairedSecretShare<Normal>, KeyPair)> = dual_role_receivers
+        .into_iter()
+        .map(|(receiver, enckey)| {
+            let CertifiedSecretShare { paired_share, .. } = receiver
+                .finalize(&cert_scheme, cert_map.clone(), &contributor_public_keys)
+                .expect("simulate_keygen finalize");
+            (paired_share.non_zero().unwrap(), enckey)
+        })
+        .collect();
 
     SimulatedKeygenOutput {
         certified_keygen,
--- a/schnorr_fun/src/frost/chilldkg/certpedpop.rs
+++ b/schnorr_fun/src/frost/chilldkg/certpedpop.rs
@@ -13,6 +13,7 @@ pub mod certificate;
 pub use certificate::vrf_cert;
 pub use certificate::{
     CertificateError, CertificationScheme, CertifiedKeygen, Certifier, CertifierError,
+    RecoverShareError,
 };
 
 use super::{encpedpop, simplepedpop};
--- a/schnorr_fun/src/frost/chilldkg/certpedpop.rs
+++ b/schnorr_fun/src/frost/chilldkg/certpedpop.rs
@@ -56,6 +56,7 @@ impl Contributor {
     pub fn gen_keygen_input<H: Hash32, NG: NonceGen>(
         schnorr: &Schnorr<H, NG>,
         threshold: u32,
+        n_contributors: u32,
         receiver_encryption_keys: &BTreeMap<ShareIndex, Point>,
         my_index: u32,
         rng: &mut impl rand_core::RngCore,
@@ -63,6 +64,7 @@ impl Contributor {
         let (inner, message) = encpedpop::Contributor::gen_keygen_input(
             schnorr,
             threshold,
+            n_contributors,
             receiver_encryption_keys,
             my_index,
             rng,
@@ -248,7 +250,7 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
         .map(|(party_index, enckeypair)| (*party_index, enckeypair.public_key()))
         .collect::<BTreeMap<ShareIndex, Point>>();
 
-    let n_generators = n_receivers + n_extra_generators;
+    let n_contributors = n_receivers + n_extra_generators;
 
     // Generate keypairs for contributors - receivers will use their existing keypairs
     let contributor_keys: Vec<_> = (1..=n_receivers)
@@ -263,9 +265,16 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
         .collect();
 
     let (contributors, to_coordinator_messages): (Vec<Contributor>, Vec<KeygenInput>) = (0
-        ..n_generators)
+        ..n_contributors)
         .map(|i| {
-            Contributor::gen_keygen_input(schnorr, threshold, &public_receiver_enckeys, i, rng)
+            Contributor::gen_keygen_input(
+                schnorr,
+                threshold,
+                n_contributors,
+                &public_receiver_enckeys,
+                i,
+                rng,
+            )
         })
         .unzip();
 
@@ -274,7 +283,7 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
         .map(|kp| kp.public_key())
         .collect::<Vec<_>>();
 
-    let mut aggregator = Coordinator::new(threshold, n_generators, &public_receiver_enckeys);
+    let mut aggregator = Coordinator::new(threshold, n_contributors, &public_receiver_enckeys);
 
     for (i, to_coordinator_message) in to_coordinator_messages.into_iter().enumerate() {
         aggregator
@@ -290,7 +299,8 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
         cert_scheme.clone(),
         agg_input.clone(),
         &contributor_public_keys,
-    );
+    )
+    .expect("simulate_keygen produces matching contributor counts");
 
     let mut paired_secret_shares = vec![];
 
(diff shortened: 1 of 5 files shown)


schnorr_fun/src/frost/chilldkg/simplepedpop.rs
```diff
@@ -15,6 +15,8 @@ use alloc::{
 use core::num::NonZeroU32;
 use secp256kfun::{KeyPair, hash::Hash32, nonce::NonceGen, poly, prelude::*, rand_core};
 
+const POP_DOMAIN_SEP: &str = "BIP DKG/pop";
```
nickfarrow: The draft bip uses `BIP DKG/pop message`
https://github.com/BlockstreamResearch/bip-frost-dkg/blob/master/python/chilldkg_ref/simplpedpop.py#L35

schnorr_fun/src/frost/chilldkg/certpedpop.rs
nickfarrow: This function verifies certificates for the expected parties in `cert_keys`, but propagates the caller-supplied certificate map which *could* contain extra `(cert_key, certificate)`s, which is then included in the VRF beacon.

I don't think this is exploitable with how we are currently using it in Frostsnap, devices & coordinator independently build their own certificate and only include known participants. But it is a rough edge that could be solved by filtering on `cert_keys` or asserting length of cert keys matches certificate map length in `finalize`.

nickfarrow (dismissed):
Noticed that b6389207dc1c9b169911099ec26a9e01607cc7cd says 
```
    - simplepedpop::receive_secret_share takes n_contributors and rejects
      length mismatches with a new ReceiveShareError::WrongContributorCount.
```
but that was not implemented.

I think this is ok since a malicious coordinator inflating the contributor count should achieve nothing given the security model. The dangerous situation would be if a coordinator were able to inflate the receiver count, in frostsnap we check this count in the keygen check.

schnorr_fun/src/frost/chilldkg/certpedpop/certificate.rs
```diff
@@ -73,11 +73,11 @@ impl<Sig> CertifiedKeygen<Sig> {
...
-    /// Verify that all certificates are valid for the given certification scheme and contributor keys.
+    /// Re-verify every stored certificate against the scheme and contributor keys.
     ///
-    /// This type should normally be impossible to construct in an invalid state through the API,
-    /// but since it supports serialization, this method allows validating instances that have been
-    /// deserialized or received from untrusted sources.
+    /// `CertifiedKeygen` can only be constructed via [`Certifier::finish`], which
+    /// verifies each certificate as it is received, so this method is a sanity check
+    /// rather than a safety gate.
     pub fn verify<S>(
```
LLFourn, line 81: eh we don't need a "sanity check" polluting the API. We could turn this into simpler `new` function which just takes a list of certificiates.

schnorr_fun/src/frost/chilldkg/certpedpop/certificate.rs
```diff
@@ -290,12 +290,21 @@ pub struct Certifier<S: CertificationScheme> {
...
+    /// Returns [`CertifierError::ContributorCountMismatch`] if the `agg_input` claims
+    /// a different number of contributors than `contributor_keys.len()`.
     pub fn new(
         cert_scheme: S,
         agg_input: encpedpop::AggKeygenInput,
         contributor_keys: &[Point],
-    ) -> Self {
+    ) -> Result<Self, CertifierError> {
+        if agg_input.n_encryption_nonces() != contributor_keys.len()
+            || agg_input.inner().n_contributors() != contributor_keys.len()
```
LLFourn, line 316: feels like the n_contributors and n_encryption_nonces should already be equal here. This is checking the validity of agg_input implicitly but it should already be correct. agg_input should just have a method `n_contributors()` so you don't have to call `inner()`.

In order to enforce the invariant correctly we'll have to make WireAggkeygenInput that is serializable and has an invariant checker that converts it to AggKeygenInput.

schnorr_fun/src/frost/chilldkg/certpedpop/certificate.rs
```diff
@@ -398,6 +399,9 @@ pub enum CertifierError {
     InvalidSignature,
     /// Not all required certificates have been received
     IncompleteCertificates,
+    /// The aggregated keygen input has a different number of contributors than
+    /// the provided `contributor_keys` list.
+    ContributorCountMismatch,
```
LLFourn, line 417: Seems to me that this error variant is only used in one of the functions and is no tused in receive_certificate even though it returns CertifierError. We should make the errors tight.

schnorr_fun/src/frost/chilldkg/certpedpop.rs
```diff
@@ -336,6 +374,18 @@ pub fn simulate_keygen<H: Hash32, NG: NonceGen, S: CertificationScheme + Clone>(
         .finish()
         .expect("Certifier should have all required certificates");
 
+    // Now that certification is done, release each dual-role party's share.
+    let cert_map = certified_keygen.certificate().clone();
+    let paired_secret_shares: Vec<(PairedSecretShare<Normal>, KeyPair)> = dual_role_receivers
+        .into_iter()
+        .map(|(receiver, enckey)| {
+            let CertifiedSecretShare { paired_share, .. } = receiver
```
LLFourn, line 379: shouldn't we check that the certified keygen obtained from this is eq to the one above.

schnorr_fun/src/frost/chilldkg/encpedpop.rs
```diff
@@ -132,6 +146,16 @@ pub struct AggKeygenInput {
...
 impl AggKeygenInput {
+    /// The inner simplepedpop aggregated input.
+    pub fn inner(&self) -> &simplepedpop::AggKeygenInput {
+        &self.inner
+    }
+
+    /// The number of encryption nonces in this aggregated input (one per contributor).
+    pub fn n_encryption_nonces(&self) -> usize {
+        self.encryption_nonces.len()
+    }
```
LLFourn, line 152: I don't think we should have this API -- it should just be `n_contributors` and call down to `simplepedpop` as stated before.

schnorr_fun/src/frost/chilldkg/simplepedpop.rs
```diff
@@ -386,6 +425,10 @@ pub enum ReceiveShareError {
     InvalidPop,
     /// The secret share we got was invalid
     InvalidSecretShare,
+    /// No encrypted share exists at the receiver's share index.
+    UnknownShareIndex,
+    /// The supplied keypair isn't the encryption key registered for this share.
+    WrongEncryptionKey,
```
LLFourn, line 431: these are defined here but never used in this file. "EncryptionKey" should be a clue. They should be defined in encpedpop.

schnorr_fun/src/frost/chilldkg/encpedpop.rs
```diff
@@ -131,7 +152,65 @@ pub struct AggKeygenInput {
...
+        let wire = WireAggKeygenInput::deserialize(deserializer)?;
+        wire.into_validated()
+            .map_err(crate::fun::serde::de::Error::custom)
+    }
+}
+
 impl AggKeygenInput {
+    /// The number of contributors in this aggregated input.
+    pub fn n_contributors(&self) -> usize {
+        self.encryption_nonces.len()
```
LLFourn, line 156: This should call simplepedpop's AggKeygenInput's n_contributors.

schnorr_fun/src/frost/chilldkg/simplepedpop.rs
```diff
@@ -402,8 +441,71 @@ impl core::fmt::Display for ReceiveShareError {
...
+
+/// Build the bytes that the proof-of-possession signs.
+///
+/// Binding format is `parity_byte || contributor_index_be`, where `parity_byte`
+/// is `0` if `com[0]` has even y and `1` otherwise. Including the parity closes
+/// the BIP340 x-only gap: a replay of `(A, pop)` under the negated key `-A`
+/// reconstructs a different message and so fails PoP verification.
+///
+/// SPEC DEVIATION: we add the parity byte where as the spec only has the contribution_index.
+fn pop_message_bytes(first_coeff: Point, contributor_index: u32) -> [u8; 5] {
```
LLFourn, line 507: I think we should just put the is_event_y as the first param to avoid the impression we're rehashing the whole point.

LLFourn:
@nickfarrow thanks for catching the stale commit message. I probably asked it to undo it because it wasn't doing anything. Effectively, trying to tighten security so that much more things are checked automatically was a bit of an odessy. The current implementation has a very loose API and relies on the checking symmetric view accross parties for security. Since security always rests on that check this is not really wrong but this API could do with a lot more defense in-depth. There were various and subtle ways of using it insecurely most of which we avoided in frostsnap but it would be much more confident if the API didn't have any.

I built #248 on top of this one but it might be better just to review #248 in its entireity rather than continueing tto review this one.  The purpose of #248 is to create a less flexible but absoluitely impossible to fuck up API.

nickfarrow (approved):

PR #248, opened 2026-04-30
chilldkg: tighten API and internal representation

LLFourn:
## Summary

Stacked on #247. A single squashed commit that tightens the chilldkg API.

The core idea: each `Contributor` now saves the parameters it was set up with (encryption keys, contributor nonce, threshold, role, slot) and verifies the aggregated input against that saved view rather than trusting whatever the coordinator returns. `cert_bytes()` incorporates the saved keysets, so a malicious coordinator that shows different parties different views of the receiver- or aux-keyset produces different `cert_bytes` per victim and mutual certification fails.

The biggest consequences:

- **Receiver encryption keys leave the encpedpop wire form** (they were embedded next to each encrypted share) and **aux contributor keys stop being a `finalize` parameter**; both move onto the contributor at `gen_keygen_input` time and feed into `cert_bytes()` via the verified state.

- **`encpedpop::AggKeygenInput` length consistency is enforced on deserialize** via a private wire type, so downstream code can assume the embedded simplepedpop contributor count matches the encryption-nonce count.

- **`Contributor` is type-parameterized by role** (`ShareReceiver` | `AuxContributor`). The role decides slot-index translation at compile time and selects the verify shape; `ShareReceiver`'s verify atomically pairs the secret share, and the share is gated behind a final certificate check.

- **Errors are split into per-function variants** so each call site enumerates exactly the conditions it can produce; an `EncryptionCheckError` sub-enum is shared between aux and share-receiver verify paths so simplepedpop errors flow through one `Inner` path with no ambiguity.

The only thing missing from this PR is a way to "keep around" the final agg keygen so you can actually recover from it. We don't use that in frostsnap but the feature has been soft-deleted since it's hard to create a `VerifiedAggKeygenInput` (which is the thing you can actually recover from) and you can't serialize it.

Frostsnap is updated against this branch in [frostsnap/frostsnap#chilldkg-api-update](https://github.com/frostsnap/frostsnap/tree/chilldkg-api-update).

### DIFF
--- a/schnorr_fun/src/frost/chilldkg/certpedpop.rs
+++ b/schnorr_fun/src/frost/chilldkg/certpedpop.rs
@@ -1,22 +1,22 @@
 //! `certpedpop` is built on top of [`encpedpop`] to add certification of the outcome.
 //!
-//! In [`encpedpop`] and [`simplepedpop`] it's left up to the application to figure out whether all
-//! the parties agree on the `AggKeygenInput`. In `certpedpop` the relevant methods return
-//! certification signatures on `AggKeygenInput` once they've been validated so they can be
-//! collected by the share receivers. Once the share receivers have got all the certificates they
-//! can finally output the key.
-//!
-//! Certificates are collected from other share receivers as well as `Contributor`s.
+//! At the simplepedpop and encpedpop layers, confirming every party has the same
+//! [`AggKeygenInput`] is the application's problem. `certpedpop` packages it: each
+//! party's verify produces a certificate signature (collected via [`Certifier`])
+//! and a receiver's secret share is withheld in [`SecretShareReceiver`] until every
+//! certificate is in.
 
 pub mod certificate;
 #[cfg(feature = "vrf_cert_keygen")]
 pub use certificate::vrf_cert;
 pub use certificate::{
-    CertificateError, CertificationScheme, CertifiedKeygen, Certifier, CertifierError,
-    RecoverShareError,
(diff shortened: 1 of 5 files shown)


schnorr_fun/src/frost/chilldkg/certpedpop/certificate.rs
```diff
@@ -16,24 +16,53 @@ pub trait CertificationScheme {
...
+    fn verify_cert(
+        &self,
+        cert_key: Point,
+        verified_input: &VerifiedAggKeygenInput,
+        signature: &Self::Signature,
+    ) -> bool {
+        (**self).verify_cert(cert_key, verified_input, signature)
+    }
+}
+
```
LLFourn: I wonder why we needed this.

schnorr_fun/src/frost/chilldkg/certpedpop.rs
```diff
@@ -1,411 +1,454 @@
...
+        let my_aux_index = self.inner.role_index() as usize;
+        let Self {
+            inner,
+            aux_contributor_keys,
+            ..
+        } = self;
+
+        let inner_verified = inner.verify_agg_input(pop_schnorr, agg_input)?;
+        let expected_cert_key = aux_contributor_keys[my_aux_index];
+        if expected_cert_key != keypair.public_key() {
```
nickfarrow, line 105: Tiny nit: we do the expensive `verify_agg_input` before this check

nickfarrow (dismissed):

nickfarrow (approved):

